PR #2, opened 2016-09-06
OS X: Do not build an .app bundle for CLI tools

thp:
With QMake 2.01a from Qt 4.8.7 (installed via Homebrew), the current codebase built ".app" bundles for `himdcli` and `netmdcli`. With this change, no app bundle is created, and the tools are created as normal binaries.


glaubitz:
Makes sense, thanks for spotting this!

I'm going to pick up development again in the near future, btw. So if you're interested contributing, just let me know.

PR #3, opened 2016-09-06
Remove unused virtual functions / use pure virtual functions

thp:
Removed some dead code that was never reached. As there are only two known subclasses (NetMD- and HiMD-specific) for each of these classes, there's no need for the base class to provide a dummy implementation.

### DIFF
--- a/qhimdtransfer/qmddevice.h
+++ b/qhimdtransfer/qmddevice.h
@@ -48,12 +48,11 @@ public:
     virtual void * deviceHandle();
     virtual void registerMdChange(void * regMdChange);
     virtual void * MdChange();
-    virtual QMDTrack track(unsigned int trkindex) {return QMDTrack();}
     virtual int trackCount() {return trk_count;}
     virtual QStringList downloadableFileExtensions() const;
     virtual void checkfile(QString UploadDirectory, QString &filename, QString extension);
-    virtual void batchUpload(QMDTrackIndexList tlist, QString path) {}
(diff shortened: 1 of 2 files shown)


glaubitz:
Thanks for the additional PRs.

We used our own git repository in the past instead of github and patches were usually handled over the mailing list [1], but I'm perfectly fine with merging PRs on github as well, so please feel free to open additional PRs any time.

I have invited Michael Karcher (@karcherm) to be a collaborator for the linux-minidisc repository, so I can add him as a reviewer. He was one the other main developers in the project and I would like him to have a look at your changes as well.

Thanks a lot!

> [1] https://lists.fu-berlin.de/listinfo/linux-minidisc

karcherm:
I expect this pull request has been compile-tested. Looks good.

thp:
Compile-tested at least on `Apple LLVM version 7.3.0 (clang-703.0.31)`. Haven't tested on Linux/GCC at the moment -- wonder if it makes sense to set up a Travis-CI build for Linux and OS X as well as an AppVeyor CI build for Windows (Visual Studio)? This would also build pull requests automatically to check if it compiles and (if there are any) unit tests pass.

glaubitz:
Let me check out Thomas' tree and compile-test on Linux/gcc. Will do that for all the currently open PRs.

glaubitz:

karcherm:
I was asked to confirm again. The patch still looks good.

PR #4, opened 2016-09-06
libhimd: Add missing #include directives, fix unused parameters, cast id3tag return types

thp:
This avoids compiler warnings with recent versions of clang.

### DIFF
--- a/libhimd/himd.c
+++ b/libhimd/himd.c
@@ -31,7 +31,9 @@
 
 #define G_LOG_DOMAIN "HiMD"
 #include <glib.h>
+#include <glib/gstdio.h>
 #include "himd.h"
+#include "himd_private.h"
 
 #define _(x) (x)
 
--- a/libhimd/himd.c
+++ b/libhimd/himd.c
@@ -165,7 +165,7 @@ int himd_write_tifdata(struct himd * himd, struct himderrinfo * status)
     gchar *filepath;
     GDir * dir;
     GError * error = NULL;
-    status = status;
+    (void)status;
 
     filepath = g_build_filename(himd->rootpath,himd->need_lowercase ? "hmdhifi" : "HMDHIFI", NULL);
     dir      = g_dir_open(filepath,0,&error);
--- a/libhimd/mdstream.c
+++ b/libhimd/mdstream.c
@@ -250,8 +250,8 @@ int himd_writestream_write(struct himd_writestream * stream, struct blockinfo * 
     unsigned char data[HIMD_BLOCKINFO_SIZE];
     g_return_val_if_fail(stream != NULL, -1);
     g_return_val_if_fail(audioblock != NULL, -1);
-    status = status;
-    stream = stream;
+
+    (void)status;
 
     // serialize the block descriptor
     setblock(audioblock, data);
--- a/libhimd/mp3tools.c
+++ b/libhimd/mp3tools.c
@@ -25,6 +25,7 @@
 
 #include <id3tag.h>
 #include "himd.h"
+#include "himd_private.h"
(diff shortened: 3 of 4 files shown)


karcherm:
The first commit looks OK. The title is misleading: I would expect that "Fix return type of ..." would change the definition of "...". In this case, we are casting the return values instead. The casts are correct, as id3 decided to use its own "utf8" character type (which is defined "signed char"), while we use (as glib does) plain char. I am slightly surprised that this is an issue in C. I know that in C++, char, signed char and unsigned char are treated as three different types, but I used to think that in C (and I am quite confident it is that way in C89) that char is an alias for either signed char or unsigned char (and not a type on its own). So the cast from signed char to char should be unneeded in C.
If a compiler gets picky about it, the cast is OK, but I suggest to refactor it into a subfunction instead, like

```
char* dup_id3_first_string(const union mp3_field *field)
{
    return id3_ucs4_utf8duplicate(id3_field_getstrings(field,0));
}
```

</code>

karcherm:
Handling unused parameters by casting them to void instead of self-assignment is a good idea, yet there are a couple of things I dislike in the second commit:
libhimd is (AFAIK) clean C89 with variable declarations at the top of the blocks before any code. I don't like to give up that property for just casting parameters to void before declaring variables. I'd rather put the void casts just where the self-assignments are now.
libhimd uses the glib layer on top of C89. glib provides the macro G_GNUC_UNUSED, which should work on clang, despite its name. One should consider using G_GNUC_UNUSED in the function declaration instead of either void-casting or self-assignment. I consider both void-casting and G_GNUC_UNUSED as viable alternatives at the moment.

The unused status parameter in himd_write_tifdata points to a deeper issue in that code. The error handling in that function using exit(1) is deeply broken. If I remember correctly, the reason to accept that code at all was that the error paths should be unreachable anyway, as this function may not be called unless other functions already ensured that there is a track index file, and that the disc is not write protected. We should nevertheless return a valid status instead of exiting in that case. If this is fixed, we don't need to handle the unusedness of the status parameter.

In himd_writestream_write, stream is not unused at all. We don't need either `(void)stream` or `stream = stream`. For status handling, the same is valid as in himd_write_tifdata.

For `himd_add_track_info` and  `himd_fragment_track_info`, the changes look OK for now, although I strongly suspect that we need to handle the "out of fragment descriptors" case.

So, summary on what I strongly suggest to change before merging:
1) Leave the statements to handle unused parameters where they are _or_ use G_GNUC_UNUSED in the parameter list of the implementation of the function
2) Don't use any workaround  for unused parameters on stream in himd_writestream_write.

Furthermore, I like to have addressed (maybe separately):
3) Handle errors more gracefully.

karcherm:
(further comments later this day)

karcherm:
The last commit about the include files is perfect.

thp:
Fixed up the commit with the void casts (moved below variable declarations, removed the void-cast for the one situation where the variable is used in the function).

As mentioned in one of the other pull requests, not sure why C89 is a practical/technical goal.

For unreachable code, GLib provides `g_assert_not_reached()` or `g_error()` - those (can) produce a core dump, which might be more useful than `exit()`, especially in situations where we don't expect the code to end up there. In any case, once the bug is fixed, we can remove the void-cast.

Also fixed up the id3tag commit, and added the `dup_id3_first_string()` as suggested.

glaubitz:

karcherm:
The patch looks fine now. With graceful error handling, I din't mean `g_assert_not_reached()` or something like that, but to use `set_status_printf` and returning an error code. Anyway, this is stuff for a different patch.

PR #5, opened 2016-09-06
libnetmd: Compiler fixes

thp:
Handle type mismatches, uninitialized variables (bail out on unknown disk format) and format strings on 64-bit systems.

### DIFF
--- a/libnetmd/libnetmd.c
+++ b/libnetmd/libnetmd.c
@@ -175,7 +175,7 @@ int netmd_set_title(netmd_dev_handle* dev, const uint16_t track, const char* con
     int oldsize;
 
     /* the title update command wants to now how many bytes to replace */
-    oldsize = netmd_request_title(dev, track, reply, sizeof reply);
+    oldsize = netmd_request_title(dev, track, (char *)reply, sizeof(reply));
     if(oldsize == -1)
         oldsize = 0; /* Reading failed -> no title at all, replace 0 bytes */
 
@@ -476,7 +476,7 @@ int netmd_set_disc_title(netmd_dev_handle* dev, char* title, size_t title_length
     int oldsize;
 
     /* the title update command wants to now how many bytes to replace */
-    oldsize = request_disc_title(dev, reply, sizeof reply);
+    oldsize = request_disc_title(dev, (char *)reply, sizeof(reply));
     if(oldsize == -1)
         oldsize = 0; /* Reading failed -> no title at all, replace 0 bytes */
 
--- a/libnetmd/libnetmd.c
+++ b/libnetmd/libnetmd.c
@@ -956,7 +956,7 @@ int netmd_write_track(netmd_dev_handle* devh, char* szFile)
     read(fd,data,4);
     data_size_i = (size_t)(data[3] + (data[2] << 8) + (data[1] << 16) + (data[0] << 24));
 
-    fprintf(stderr,"Size of data: %d\n",data_size_i);
+    fprintf(stderr,"Size of data: %lu\n", (unsigned long)data_size_i);
     size = (data_size_i/0x3f18)*8+data_size_i+8;           /* plus number of data headers */
     fprintf(stderr,"Size of data w/ headers: %d\n",size);
 
@@ -990,7 +990,9 @@ int netmd_write_track(netmd_dev_handle* devh, char* szFile)
 
         file_pos = (size_t)lseek(fd,0,SEEK_CUR); /* Gets the position in file, might be a nicer way of doing this */
 
-        fprintf(stderr,"pos: %d/%d; remain data: %d\n", file_pos, data_size_i, data_size_i-file_pos);
+        fprintf(stderr,"pos: %lu/%lu; remain data: %ld\n",
+                (unsigned long)file_pos, (unsigned long)data_size_i,
+                (signed long)(data_size_i - file_pos));
         if (file_pos >= data_size_i) {
             fprintf(stderr,"Done transferring.\n");
             free(data);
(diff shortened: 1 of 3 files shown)


karcherm:
All patches look quite good. The format string patch introduce a dependency on C99-like format strings, which is no problem for gcc and clang (as glibc supports %zd for ages), but requires Microsoft Visual C++ from Visual Studio 2015 or newer. glib has G_GSIZE_MODIFIER, and I would prefer to use that until no one in the Windows community uses VS2013 (or older) anymore. So instead of `"Size of data: %zu\n"` i prefer `"Size of data: %" G_GSIZE_MODIFIER "u"` for greater portability.

While I personally prefer sizeof for named objects without the parentheses, I noticed that the remaining sizeof invocations in libnetmd.c include parentheses, so adding those parentheses to `sizeof(reply)` is the right thing to do, as it improves consistency.

karcherm:
I'm sorry, I mixed things up. The patches are fine as-is. libnetmd (opposed to libhimd) is not based on glib, and I don't like introducing glib just for getting G_GSIZE_MODIFIER.

thp:
Is there a technical reason for wanting to support VS2013 or C89? Is this version of libnetmd used outside the linux-minidisc project? Even if it is, having a build-requirement of VS2015 and a C99 compiler (if not C11/C++11) seems like a reasonable thing in 2016, especially for a self-contained project such as this. VS2015 Community Edition is free for "non-enterprise applications", which this one clearly is, and VS2015 can still target Windows XP if needed (not to mention MinGW as possible alternative toolchain).

glaubitz:

karcherm:
My experience with windows development is that people often use older versions of Visual Studio, most likely because of legacy code or laziness regarding upgrades. For example, Wireshark still targets Visual Studio 2010 by default.
The Windows Builds for QHiMDTransfer are built using MinGW. If I googled correctly, MinGW supports %zd only if __USE_MINGW_ANSI_STDIO is defined. I am quite confident we currently _don't_ do that in linux-minidisc, except if qmake adds it automatically. This means we might need to add this to the qmake project definition to avoid breaking the MinGW executables. This should be verified.

thp:
Yes, I think i saw a warning about `%z` when cross-compiling with MinGW this weekend. So would it be better to use `%lu` and `%ld` then and cast to `unsigned long` and `long` instead, or should we just define that macro (not sure what other side effects that macro has)?

karcherm:
Mingw has its own ANSI-conformant stdio implementation, but uses the stdio implementation from msvcrt.dll by default. So MinGW supports the typical windows formatting extensions (like "%I64d"), but not the C99 stuff. It is likely that if you mix MinGW and MSVCRT stdio stuff, you get two independent buffers, so printing order might get messed up. For application programs (like netmdcli), we have full control over what code that prints stuff is invoked (most likely just our own code), so interoperability with code that uses MS stdio is not needed, and defineing __USE_MINGW_ANSI_STDIO is fine. On the other hand, in libnetmd I am not really fond of definig this macro, as libnetmd might get linked into code we don't know yet, and I don't want to force __USE_MINGW_ANSI_STDIO on libnetmd users.
On a second glance, it only seems to apply to debug messages that (unconditionally?!) are printed to stderr. stderr is supposed to be line-buffered at most, and we print full lines, so the buffering stuff shouldn't be an issue.
I suggest to define the macro and go on with C99 format strings in this case.

thp:
According to this Wiki page, it might not be so simple (in terms of silencing the compiler warnings), and I don't really want to define a custom `my_print` as suggested there:

https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/

Would you be happy with casting to long instead? For netmd, it's probably not like we're going to have byte positions > 4.294.967.296 (or 2.147.483.648 in the signed case), and it's just debug output anyway, as you said.

karcherm:
I would be completely happy with casting to _unsigned_ long instead. size_t is unsigned, so it makes no sense to cast to (signed) long.

thp:
Updated pull request to use `%lu` and casting to `(unsigned long)`, except for the one case where we calculate a difference, there I used `%ld` and cast to `(signed long)` (which is the same as `long`, but more explicit in that context).

karcherm:
This is good enough to get merged.

Just a nitpick: The C standard says that when you calculate the difference between two size_t (or any other unsigned integer types that is at least as big as int) variables, the result is again unsigned. If the difference is less than zero, you get well-defined wrapping (opposed to signed types where overflow/underflow results in undefined behaviour). So if you calculate (2U - 4U) you are guaranteed to get 0xFFFFFFFE on a 32 bit system. Casting this to a signed type that can't represent 4 billion is implementation defined. In C99, the compiler writer must ensure that the conversion consistently either happens according to the rules of sign/magnitude (in that case the result would be -0x7ffffffe), two's complement (in that case the result would be -2) or one's complement (in that case the result would be -1), and the way this conversion is treated must be documented by the compiler writer. As all notable  general-purpose microprocessors today internally use two's complement, no compiler writer for any relevant platform is going to decide anything but treating the unsigned value as being a two's complement number, and in this case, the cast to signed really does what we want. To be safe according to the standard, you would have to cast to signed long before you calculate the difference, but you will be doomed if the numbers exceed 2G before casting (in which case the difference might not fit into 32 bits signed _at all_...), so this also is no universal solution. The "perfect" way would be manually distinguishing whether the result would be positive or negative and manually printing the sign like this:
    size_t a, b;
    a = foo();
    b = bar();
    if (a > b) {
        printf("%lu", static_cast<unsigned long>(a - b));
    } else {
        printf("-%lu", static_cast<unsigned long>(b - a));
    }
which would be definitely overkill here, but correctly handle the 33-Bit result you can obtain when subtracting two 32 Bit numbers. (32 Bits are passed in the numeric argument and the 33'th bit is the choice of the format string).
The difference that started the discussion (total - remaining) should always yield a positive result (except if something goes wrong or is buggy), so I would have used %lu at that place, too.

PR #6, opened 2016-09-06
netmdcli: Format string fixes, fix netmd_get_devname() API in libnetmd

thp:
Two small compiler warning fixes for netmdcli.

### DIFF
--- a/libnetmd/netmd_dev.c
+++ b/libnetmd/netmd_dev.c
@@ -130,11 +130,11 @@ netmd_error netmd_open(netmd_device *dev, netmd_dev_handle **dev_handle)
     }
 }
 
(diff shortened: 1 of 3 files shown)


karcherm:
In my oppinion, the interface of netmd_get_devname is broken. libnetmd uses char (not unsigned char) in its public interface for strings in all other places, so expecting netmd_get_devname to accept the address of an char array is a reasonable assumption. netmd_get_devname should include the cast from `char*` (libnetmd convention for strings) to `unsigned char*` (what libusb expects).
The size_t fix is fine.

thp:
Changed the `netmd_get_devname()` API instead and dropped the change in `netmdcli`.

glaubitz:

karcherm:
Looks great, can be applied.

PR #7, opened 2016-09-06
md.pro: Fix subproject dependencies / parallel builds

thp:
The `CONFIG += order` probably doesn't do anything. There's a `CONFIG += ordered`, but it's use is [discouraged](https://blog.rburchell.com/2013/10/every-time-you-configordered-kitten-dies.html). This change makes the subdirs dependencies explicit and allows proper subproject building (e.g. `make sub-netmdcli`) as well as parallel builds (e.g. `make -j8`).

### DIFF
--- a/README
+++ b/README
@@ -10,10 +10,10 @@ sox 14.2 (for the GUI)
 BUILDING
 
 This project uses the qmake build system to cooperate with the Qt Creator
-IDE. To compile in Qt Creator, just open himd.pro. To compile at the
+IDE. To compile in Qt Creator, just open md.pro. To compile at the
 command line, run
 
-  qmake -r
+  qmake
   make
 
 Depending on the default configuration of your system, you will find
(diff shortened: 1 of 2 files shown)


karcherm:
I have no idea whether building app bundles for the CLI tools is a good idea, I leave that to @glaubitz, but the fixes to README and md.pro look great. Thanks for that!

glaubitz:
We are currently building the app bundles for the CLI tools which is not what you really want for terminal applications as the binary is hidden in the .app bundle and therefore hard to access from the terminal on OS X.

This was addressed in Thomas' first PR, but I screwed up merging it and I'm waiting for him to re-open it tonight.

thp:
Rebased this branch on the master (it was done against yesterday's master where #2 was already merged, and after the master branch was reverted, the commits appeared in this pull request).

The app bundle changes are now in #9, and it changes from the current state (creating an app bundle for CLI tools) to _not_ building app bundles for CLI tools (at least on recent Qt/qmake versions on OS X, this might have been different in the past).

glaubitz:
Builds fine on Debian unstable, tested with gcc-6 and Qt5. I can also confirm that this patch fixes parallel building.

karcherm:
The dependency declarations are a real good thing to have. Thanks.

PR #8, opened 2016-09-06
qhimdtransfer: Only prepend artist if not empty

thp:
When uploading files, do not prepend the artist name if it is empty. This avoids filenames like:

`" - Title.oma"`

in favor of

`"Title.oma"`

when there is no artist name set on the track.

In the future, we might even want to make this configurable, so the user can choose to also include the track number, album name, etc...

### DIFF
--- a/qhimdtransfer/qmddevice.cpp
+++ b/qhimdtransfer/qmddevice.cpp
@@ -599,11 +599,15 @@ void QHiMDDevice::upload(unsigned int trackidx, QString path)
     QString filename, errmsg;
     QHiMDTrack track = himdTrack(trackidx);
     QString title = track.title();
+    QString artist = track.artist();
 
-    if(title.isNull())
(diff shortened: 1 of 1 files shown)


karcherm:
Good idea, the patch looks correct, too. But if you check for a non-present artist with isEmpty (as recommended by the Qt documentation), please also adjust the check for a non-present title (which currently uses isNull). I personally prefer to have the distinction between "empty string" and "no string at all", but I guess users likely wouldn't like an empty artist prefix or empty filename, even if the string is "present but empty" instead of "missing".

thp:
Updated this one to use `.isEmpty()` for both and refactored a bit to look nicer.

glaubitz:

karcherm:
This patch looks fine now. Thank you very much for all the good quality code you submitted in the pull requests!

PR #9, opened 2016-09-07
OS X: Do not build an .app bundle for CLI tools

thp:
Re-issue of https://github.com/glaubitz/linux-minidisc/pull/2


glaubitz:
Builds fine on Debian unstable, tested with gcc-6 and Qt5 (as expected in this case).

Also verified on OS X and the changes have the desired effects.

glaubitz:
Merging this now, although there are some additional build discovered on OS X, at least when using Brew instead of Macports.

PR #11, opened 2016-09-13
Replace sox dependency with custom WAV file writer

thp:
The only reason sox is dragged in as dependency is to write out WAV files. This can be easily done with custom code, and will in turn reduce the build-dependencies for cross-compilation cases.

Since we always get the same format (44100 Hz / stereo / big-endian signed 16-bit samples), writing out the samples is just a matter of byte-swapping the incoming 16-bit values, no need to convert to `sox_sample_t` (32-bit values) and then back.

I have verified with a simple LPCM recording that the resulting WAV file written out is exactly the same byte-by-byte as the WAV file written out by the old SOX-based code.

Right now, this code lives in QHiMDTransfer and uses Qt classes, but we could later move this down to `libhimd` (just using stdio file access) and use it in `himdcli` (for the `dumpnonmp3` sub-command, to write LPCM tracks as WAV files) and QHiMDTransfer.

### DIFF
--- a/README
+++ b/README
@@ -5,7 +5,6 @@ glib (for the core library)
 mad (for MP3 transfer, can be disabled)
 libmcrypt (for PCM transfer, can be disabled)
 Qt 4 (for the GUI)
-sox 14.2 (for the GUI)
 
 BUILDING
 
--- a/qhimdtransfer/main.cpp
+++ b/qhimdtransfer/main.cpp
@@ -13,7 +13,6 @@
 int main(int argc, char *argv[])
 {
     int status;
-    sox_format_init();
 
     QApplication a(argc, argv);
     QTranslator trans;
@@ -29,6 +28,5 @@ int main(int argc, char *argv[])
     QHiMDMainWindow w;
     w.show();
     status = a.exec();
-    sox_format_quit();
     return status;
 }
--- a/qhimdtransfer/qhimdmainwindow.h
+++ b/qhimdtransfer/qhimdmainwindow.h
@@ -9,10 +9,6 @@
 #include "qhimddetection.h"
 #include "qmdmodel.h"
 
-extern "C" {
-#include <sox.h>
-}
-
 namespace Ui
 {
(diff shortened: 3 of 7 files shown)


karcherm:
I don't really like reinventing the wheel again. I don't know how many wave file writing implementations I already have on a Linux system with some sound applications installed, and this just adds another one. On the other hand, the code looks clean and correct, and sox is oversized for just writing wave files. Converting from 16 bit to 32 bits and back to 16 bits, which is necessary when using sox, also doesn't make sense. So I have no strong objections about merging this.
@glaubitz please decide whether this should be merged.

glaubitz:
Hmm, I'm a bit undecided on this. The idea is good in principle, but I'm not sure whether we might need sox in the future for other things. I will have to look at the code in detail again and what we are currently using sox for and if there is any chance we might extend functionality in this regard in the future.

tharp:
As netmd download support is limited to PCM and Atrac LP files, i think we could use sox in the future to add conversion routines for other audio formats.

thp:
You can still use sox in the future if you find it's needed for anything (and somebody actually implements the features that would use it), even if you merge this.

@glaubitz This change makes WAV file writing more efficient and reduces the dependencies for cross-compilation. Sox isn't used anywhere else in the repository, just for WAV file writing (if you don't count the `sox` command-line utility used for analog uploads from netmd in the Python scripts in `netmd/`, which obviously isn't affected by this pull request).

@tharp This is about QHiMDTransfer, which doesn't have any netmd support at all at the moment. Any non-LPCM data on (Hi-)MD has been lossily compressed, and I'd rather the tool doesn't do any conversion on that data, since converting lossy formats to other lossy formats results in quality loss, and converting from lossy formats to PCM/WAV would need ffmpeg/libav, sox can't convert ATRAC3(plus) by itself.

glaubitz:
@thp Ok, I agree with this stance and I agree with @karcherm observations. Let's merge this.

PR #12, opened 2016-09-13
Replace taglib dependency with simple ID3v1 writer

thp:
Taglib was an undocumented dependency for QHiMDTransfer. It was used to tag MP3 files transferred (uploaded, i.e. MD→PC) from Hi-MD devices. Similar to #11, this makes it easier to cross-compile the code by removing a build-dependency that would have to be built and shipped otherwise (also, other code already uses libid3tag for reading MP3 tags).

While taglib presumably wrote both ID3v1 and ID3v2 tags, this solution only writes a simple ID3v1 tag, but for data retrieved from the Hi-MD, that's probably good enough.

I also decided to remove the tag comment `"*** imported from HiMD via QHiMDTransfer ***"` while I'm at it, since it's a bit annoying to have that attached to all files imported - this is not an oversight, but done on purpose ;)

This probably conflicts with #11, since both modify the .pro file and the include section at the top of qmddevice.cpp, but it's easy to fix that, since both changes just remove the pkg-config dependency and the header file inclusions of the respective libraries to be removed.

### DIFF
--- a/qhimdtransfer/qhimdtransfer.pro
+++ b/qhimdtransfer/qhimdtransfer.pro
@@ -76,8 +76,7 @@ SOURCES += main.cpp \
 win32:SOURCES += qhimdwindetection.cpp
 else:SOURCES += qhimddummydetection.cpp
 RESOURCES += icons.qrc
-PKGCONFIG += sox \
-    taglib
+PKGCONFIG += sox
 win32:LIBS += -lsetupapi \
     -lcfgmgr32
 win32:RC_FILE = qhimdtransfer.rc
--- a/qhimdtransfer/qmddevice.cpp
+++ b/qhimdtransfer/qmddevice.cpp
@@ -2,10 +2,6 @@
 #include <QMessageBox>
 #include <QApplication>
 #include <QFile>
-#include <tlist.h>
-#include <fileref.h>
-#include <tfile.h>
-#include <tag.h>
 
 extern "C" {
 #include <sox.h>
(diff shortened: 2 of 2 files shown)


karcherm:
I do think that ID3v1 only is a significant loss for Hi-MD exports. MD stuff is used a lot in Japan, and Japanese characters are unlikely to be representable in latin1.

glaubitz:
I agree with Michael here. If abandoning taglib means that we can no longer support encodings like ShiftJIS which the NetMD/HiMD devices actually support, we won't be merging this change.

I have lots of Japanese music myself that I have transferred to my HiMD Walkman in the past and I guess with this change applied, I would only see garbaged description in the Walkman's display.

tharp:
Taglib is powerful and extensible in "library user mode", maybe it can also be used to read/write atrac headers (.aea, .omg, ..) in the future.

thp:
This only affects MP3s transferred from the Hi-MD device to the host computer via QHiMDTransfer.

MP3s transferred from the Hi-MD device to the host computer via `himdcli dumpmp3` do not have any tags written at all.

For uploading MP3s from the computer to the Hi-MD device, libid3tag is used from within libhimd (`himd_get_songinfo`), but only from `himdcli`, not from QHiMDTransfer.

Ideally the whole tag writing and reading should probably be made part of `libhimd`, anyway, so uploads and downloads via `himdcli` and QHiMdTransfer are both consistent.

So the only case where this could be affected is MP3 files sent to Hi-MD via SonicStage or `himdcli` with Japanese characters and transferred from Hi-MD to the computer via QHiMDTransfer.

glaubitz:
> Ideally the whole tag writing and reading should probably be made part of libhimd, anyway, so uploads and downloads via himdcli and QHiMdTransfer are both consistent.

I agree, that would be a much better design.

> So the only case where this could be affected is MP3 files sent to Hi-MD via SonicStage or himdcli with Japanese characters and transferred from Hi-MD to the computer via QHiMDTransfer.

And that's exactly one of the use cases I have (and probably also most users in Japan), so I rather don't want to switch to a tag writer which does not support all the character encodings that the HiMD devices support.

thp:
I'll probably rewrite this PR to use taglib inside libhimd then. Can you in the meantime maybe provide a simple test case (Hi-MD folder structure as well as tagged MP3 file) with a CC-licensed MP3 file with Japanese characters, so that we can then test if uploading and downloading of MP3s with Japanese-character ID3v2 tags works properly?

glaubitz:
> Can you in the meantime maybe provide a simple test case (Hi-MD folder structure as well as tagged MP3 file) with a CC-licensed MP3 file with Japanese characters, so that we can then test if uploading and downloading of MP3s with Japanese-character ID3v2 tags works properly?

No problem, we actually already have that already available on the wiki, see:

> https://wiki.physik.fu-berlin.de/linux-minidisc/doku.php?id=himddiskformat

thp:
Haven't had the time to work on this yet, so created #20 as reminder and closing this for now.

PR #13, opened 2016-09-13
Determine build date based on host os, not target

thp:
The `date` command is run at QMake/build time. In the cross-compilation (e.g. compiling Windows binaries on Linux) case, `win32` will be true, but the system on which `qmake` runs will be Linux.

Explicitly check whether QMake is running on Windows, and assume Unix in any other case, so that Windows binaries built on Linux will use the Unix variant of calling `date` at build time.


karcherm:
Thank you for this patch, it looks completely correct!

PR #15, opened 2016-09-13
Show image selector when image is chosen

thp:
There is no point in showing an error message to the user and instructing them to click a button. Instead, act as if the user has pressed the connect button and show the folder selector.


karcherm:
The patch looks fine. This should make the user interface easier to use. Merging it.

PR #17, opened 2016-09-23
Detect mounted Hi-MD devices on OS X on startup

thp:
This only does the device detection on application startup and doesn't monitor device removals/additions like the Windows-based detection code does, but it's better than not doing any detection.

### DIFF
--- a/qhimdtransfer/qhimdtransfer.pro
+++ b/qhimdtransfer/qhimdtransfer.pro
@@ -77,8 +77,15 @@ SOURCES += main.cpp \
     qmdmodel.cpp \
     qmdtrack.cpp \
     qmddevice.cpp
-win32:SOURCES += qhimdwindetection.cpp
-else:SOURCES += qhimddummydetection.cpp
+
+win32 {
+  SOURCES += qhimdwindetection.cpp
+} else:mac {
+  SOURCES += qhimdmacdetection.cpp
+} else {
+  SOURCES += qhimddummydetection.cpp
+}
+
 RESOURCES += icons.qrc
 PKGCONFIG += taglib
 win32:LIBS += -lsetupapi \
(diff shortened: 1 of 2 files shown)


qhimdtransfer/qhimdtransfer.pro
```diff
@@ -77,8 +77,15 @@ SOURCES += main.cpp \
     qmdmodel.cpp \
     qmdtrack.cpp \
     qmddevice.cpp
-win32:SOURCES += qhimdwindetection.cpp
-else:SOURCES += qhimddummydetection.cpp
+
+win32 {
+  SOURCES += qhimdwindetection.cpp
+} else:mac {
```
glaubitz, line 83: Is the colon after the "else" actually necessary. Seems like you can just put a whitespace between the "else" and "mac" keywords.

glaubitz:
Looks good to me. Would be nice, of course, if this could be extended in the future to be on par with the functionality we provided on Windows for detecting HiMD devices.

@karcherm What do you think?

qhimdtransfer/qhimdtransfer.pro
thp: I based this on http://doc.qt.io/qt-4.8/qmake-advanced-usage.html. I think it's necessary, because otherwise it might use the `else` branch of `mac` if `win32` is set but `mac` is not (or in other words, this makes sure that the `else` branch at the end is only taken if both `!win32` and `!mac`), and thereby for the `win32` build ending up with both `qhimdwindetection.cpp` and `qhimddummydetection.cpp`, and the build failing.

glaubitz: Indeed. It's a bit confusing, but it looks you're right after reading the documentation. Thanks for pointing me at it.

glaubitz:
Ok, after looking through the changes, I'm confident to merge this without @karcherm having another look.

PR #19, opened 2016-09-29
libnetmd chunksize fix

Lt-Henry:
- Using a smaller chunksize (~1Mb)
- Fixed a missing 'else' scope which prevented trackdata to be splited in several packets

### DIFF
--- a/libnetmd/secure.c
+++ b/libnetmd/secure.c
@@ -381,7 +381,10 @@ void netmd_transfer_song_packets(netmd_dev_handle *dev,
         if (error >= 0) {
             p = p->next;
         }
-        break;
+        else {
+            /* error transfering data */
+            break;
+        }
     }
 }
 
@@ -391,7 +394,7 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
                                   unsigned char *key_encryption_key)
 {
     size_t position = 0;
-    size_t chunksize = 0xffffffffU;
+    size_t chunksize = 0x000fffffU;
     netmd_track_packets *last = NULL;
     netmd_track_packets *next = NULL;
 
--- a/libnetmd/secure.c
+++ b/libnetmd/secure.c
@@ -378,13 +378,11 @@ void netmd_transfer_song_packets(netmd_dev_handle *dev,
         free(packet);
         buf = NULL;
 
-        if (error >= 0) {
-            p = p->next;
-        }
-        else {
-            /* error transfering data */
+        if (error < 0) {
             break;
         }
+        
+        p = p->next;
     }
 }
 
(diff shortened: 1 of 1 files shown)


karcherm:
The  chunksize was 0xffffffff because with the devices I have at hand (Sony MZ-RH1, Sharp IM-MT 899) did not need any chunking at all during testing. If USB bulk transfers are limited in some way (or USB host drivers or whatever), so chunking is indeed required, go on with a patch like this. I would suggest to use a chunk size that ensures whole frames to be put into a chunk, so it should be a multiple of both 2048 (the frame size for uncompressed audio) and 192 (the chunk size or ATRAC3-SP). ATRAC3-LP is automatically covered, because its frame size is half of ATRAC3-SP. So I suggest to use 0x180000 (1.5MB) or 0xC0000 (786KB) instead of 0xfffff (1MB - 1Byte).

I think the error logic would be more clear if it was written as

```
if (error < 0) {
    break;
}
p = p->next;
```

The patch you propose is definitely doing the right thing at that place, though.

Lt-Henry:
> The chunksize was 0xffffffff because with the devices I have at hand (Sony MZ-RH1, Sharp IM-MT 899) did not need any chunking at all during testing

Mine is a Mz-NH600, an entry level Hi-MD/NetMD. But after reading libusb_bulk_transfer documentation I got this:

_libusb may have to split your transfer into a number of chunks to satisfy underlying O/S requirements, meaning that the timeout may expire after the first few chunks have completed. libusb is careful not to lose any data that may have been transferred; do not assume that timeout conditions indicate a complete lack of I/O._

As I am using PCM wireformat, my single chunk is 50Mb so, it simple timeouts at 10 seconds, and it is handled as an error, but it is not, it only needs to repeat the transfer with the remaining data. 

So, this is what imho, have to be done:

1- netmd_transfer_song_packets: Send a single chunk of data and handle timeouts, as libusb already splits data.
2- netmd_prepare_packets: There is no need to split data anymore, so this function only have to generate keys and crypt the data.

If you find it is the way to go, I can give it a try.

karcherm:
Your proposal sounds proper. We could also estimate a sensible timeout for the full transfer (for ATRAC(1) I guess the transfer duration is around real time due to the MD drive speed). USB 1.1 should be perfectly able to handle the 170 KB/s. Using a shorter timeout and retrying has the advantage of being able to monitor progess, and to abort if a single call did not transfer any data at all. Do as you prefer.

Lt-Henry:
Epic fail :(

Trying to send big chunks ( ie: 24Mb)  libusb_bulk_transfer immediately fails with a -1 status and an errno=12, which It seems to be a kind of "not enough memory" error, but this is only my guess, libusb documentation has room for improvement.

`libusb: debug [submit_bulk_transfer] need 1 urbs for new transfer with length 24634392
libusb: error [submit_bulk_transfer] submiturb failed error -1 errno=12`  

So, It seems data transfers must be split and perform several bulk transfers, at least, for PCM wireformat which size are typically 20-60MB.

I will keep on working on this branch using your proposed chunksize

PR #10, opened 2016-09-10
Continous integration (Linux, macOS, Windows via Linux cross-compilation)

thp:
Here's a summary of what's changed:
- Support for Continous Integration with Travis CI (builds Linux, macOS and Windows 32/64 bit)
- Fix build date detection for cross-compilation
- Make mingw-w64 builds self-contained (build build-dependencies like libmad and gcrypt in CI)
- Remove dependency on `sox` for the GUI and replace with a simple WAV file writer
- Remove (undocumented) dependency on `taglib` for the GUI and replace with a simple ID3v1 writer
- Cleanup the qmake .pro files and split off common parts into .pri files in `build/`
- Detect mounted Hi-MD devices on macOS on startup by scanning /Volumes/ (no monitoring yet)

The build status and build log can be observed at https://travis-ci.org/thp/linux-minidisc, once this is merged, @glaubitz can create an account on travis-ci.org (basically just log in with your Github credentials), and any future pull requests will be built and checked from the original repository as well. In the future, this will allow to automatically build + upload releases to Github Releases (for example) when the master branch is tagged (see https://docs.travis-ci.com/user/deployment/releases).

PR #21, opened 2016-09-29
Restructure build system

thp:
This pull request is split off from #10 and contains only the build system changes that will later on enable the Continous Integration builds on Travis CI.


karcherm:
This pull request does not only contain build system changes, but also changes the include directives in qhimdwindetection.cpp. It seems these changes are appropriate if you use the original Microsoft header files. When that file was written, the subdirectory "ddk" was indeed needed if you (cross-)compile with mingw using the header files mingw provides.
Did that change in the mean time?

thp:
I needed to do that change so that it compiles in Ubuntu (as on Travis CI) with Mingw-W64 (which is a fork of mingw32 with support for both 32- and 64-bit). I didn't test that change with Windows compilers, though.

Here's the build logs for building with 32- and 64-bit versions of Mingw-W64 with that code change:

https://travis-ci.org/thp/linux-minidisc/jobs/163859833
https://travis-ci.org/thp/linux-minidisc/jobs/163859831

thp:
If you really want, I can split out the changes to `qhimdtransfer/qhimdwindetection.cpp` to a separate pull request, just let me know if there's anything left to do here or if this can be merged.

thp:
_bump_ - what's the status here, should I separate out the .cpp changes into a separate pull request or can we merge this as-is?

glaubitz:
Yes, those should go into a different commit, please.

thp:
Spilt off into https://github.com/glaubitz/linux-minidisc/pull/22 and rebased this one on top of it.

thp:
Merge after #22.

qhimdtransfer/qhimdwindetection.cpp
```diff
@@ -5,10 +5,10 @@
 
 #include <windows.h>
 #include <dbt.h>
-#include <setupapi.h>
 #include <initguid.h>       // needed to handle GUIDs
-#include <ddk/ntddstor.h>   // needed for handling storage devices
-#include <ddk/cfgmgr32.h>   // needed for CM_Get_Child function
+
+#include <ntddstor.h>   // needed for handling storage devices
+#include <cfgmgr32.h>   // needed for CM_Get_Child function
```
glaubitz, line 11: Why is this still in here? Wasn't this supposed to be in #22?

build/libgcrypt.pri
```diff
@@ -0,0 +1,6 @@
+!without_gcrypt {
+  LIBS += -lgcrypt
+  DEFINES += CONFIG_WITH_GCRYPT
+}
+else: !build_pass: message(You disabled mcrypt: No PCM and ATRAC transfer will be supported)
```
glaubitz, line 5: The message should mention "gcrypt", not "mcrypt". mcrypt was dropped in c5a16774418f25ff30089589c8249d3708c8335c.

glaubitz (changes requested):

qhimdtransfer/qhimdwindetection.cpp
glaubitz: Ok, this change doesn't show up in the other view. Strange.

PR #22, opened 2016-12-01
Fix include paths for Windows API

thp:


glaubitz:
Was this tested on real Windows as well or just on mingw?

From the changes, it's consisted with how the files were moved in mingw. But I'm not sure whether the same applies to actual Windows:

root@ikarus:/mnt# find . -name "cfgmgr32.h"
./Program Files (x86)/Microsoft SDKs/Windows/v7.1A/Include/cfgmgr32.h
./Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/um/cfgmgr32.h
./Program Files (x86)/Windows Kits/8.1/Include/um/cfgmgr32.h
root@ikarus:/mnt# find . -name "ntddstor.h"
./Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/shared/ntddstor.h
./Program Files (x86)/Windows Kits/8.1/Include/shared/ntddstor.h
root@ikarus:/mnt#

Or do we just want to support mingw builds now?

PR #27, opened 2016-12-21
added netmd download support in netmdcli

tharp:
I added netmd download support in netmdcli.
Supported formats are: 16bit pcm @ 44k1Hz stereo or mono and Atrac LP2/4 stored in a wav container file.

As i tested, bitrate conversion ( pcm -> Atrac LP2/4) doesn´t work this way, transfer is fine but no audio output when playing the track. Also pcm stereo -> Atrac SP mono failed with unknown error.
So i left it the way i worked on before:
16bit pcm mono -> Atrac SP mono
16bit pcm stereo -> Atrac SP stereo
LP2 -> LP2
LP4 -> LP4

Also added some more detailed debug messages, may be very useful when testing/debugging.

Successfully tested with different stereo pcm and LP2 files, LP4 and mono not tested, yet.


libnetmd/secure.c
```diff
@@ -370,8 +370,8 @@ void netmd_transfer_song_packets(netmd_dev_handle *dev,
         memcpy(buf + 8, p->iv, 8);
         memcpy(buf + 16, p->data, p->length);
 
-        /* ... send it */
-        error = libusb_bulk_transfer((libusb_device_handle*)dev, 2, packet, (int)packet_size, &transferred, 10000);
+        /* ... send it, timeout increased for larger files*/
+        error = libusb_bulk_transfer((libusb_device_handle*)dev, 2, packet, (int)packet_size, &transferred, 80000);
```
thp: Should the timeout be somehow dependent on the `packet_size`? Or asked the other way around: For which maximum file size is 80000 a good value? Can I upload a single 2-hour-track with this?

libnetmd/secure.c
```diff
@@ -418,8 +419,10 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
             chunksize = data_lenght - position;
         }
 
-        if ((chunksize % 8) != 0) {
-            chunksize = chunksize + 8 - (chunksize % 8);
+        /* do not truncate frames, transfer data size will be calculated by number of frames in netmd_secure_send_track(),
+           alternatively change totalbytes calculation in netmd_secure_send_track() */
+        if ((chunksize % frame_size) != 0) {
```
thp: This can be rewritten without a condition:

```
chunksize = ((chunksize + frame_size - 1) / frame_size) * frame_size;
```

netmdcli/netmdcli.c
```diff
@@ -168,6 +168,73 @@ void retailmac(unsigned char *rootkey, unsigned char *hostnonce,
...
+            return 0;
+        return 1;
+    }
+
+    if(leword16(file +20) == NETMD_RIFF_FORMAT_TAG_ATRAC3)   /* ATRAC3 */
+    {
+        *conversion = 0;                                     /* conversion not needed */
+        if(leword32(file+24) != 44100)                       /* sample rate */
+            return 0;
+        if(leword16(file+32) == 384) {                       /* data block size LP2 */
```
thp: Maybe make 384 and 192 a constant `NETMD_DATA_BLOCK_SIZE_LP2` and `NETMD_DATA_BLOCK_SIZE_LP4`?

netmdcli/netmdcli.c
```diff
@@ -168,6 +168,73 @@ void retailmac(unsigned char *rootkey, unsigned char *hostnonce,
     gcry_cipher_close(handle2);
 }
 
+static inline unsigned int leword32(const unsigned char * c)
+{
+    return c[3]*16777216+c[2]*65536+c[1]*256+c[0];
```
thp: Maybe easier to read:

```
return (c[3] << 24) + (c[2] << 16) + (c[1] << 8) + c[0];
```

netmdcli/netmdcli.c
```diff
@@ -168,6 +168,73 @@ void retailmac(unsigned char *rootkey, unsigned char *hostnonce,
...
+}
+
+static inline unsigned int leword16(const unsigned char * c)
+{
+    return c[1]*256+c[0];
+}
+
+static int wav_data_position(const unsigned char * data, size_t len)
```
thp: Is it possible to make that less fragile by parsing the WAV file header if possible and looking for the audio data offset that way and not scanning for the string "data"? What if some of the header fields by chance contain "data"?

netmdcli/netmdcli.c
```diff
@@ -168,6 +168,73 @@ void retailmac(unsigned char *rootkey, unsigned char *hostnonce,
...
+{
+    int pos = -1, i = 0;
+    while(pos < 0)
```
thp: The first conditional-break can be moved into the while condition, and it's actually better to use `break` when pos is set than checking for `pos < 0` in every iteration):

```
while (i < len - 4) {
    if (strncmp(...) {
        pos = i;
        break;
    }
    i += 2;
}
```

netmdcli/netmdcli.c
```diff
@@ -168,6 +168,73 @@ void retailmac(unsigned char *rootkey, unsigned char *hostnonce,
...
+        if(strncmp("data", data+i, 4) == 0)
+            pos = i;
+        i+=2;
+    }
+    return pos;
+}
+
+static int audio_supported(const unsigned char * file, netmd_wireformat * wireformat, unsigned char * diskformat, int * conversion)
+{
+    if(strncmp("RIFF", file, 4) != 0 || strncmp("WAVE", file+8, 4) != 0 || strncmp("fmt ", file+12, 4) != 0)
```
thp, line 201: This probably also needs to get the size of the data in `file` as argument and then check that the size is greater than the maximum offset we're reading here and below, otherwise we're reading out of bounds.

netmdcli/netmdcli.c
```diff
@@ -410,138 +477,192 @@ int main(int argc, char* argv[])
...
+
+                    size_t frames;
+                    int data_position, audio_data_position, audio_data_size, i, need_conversion = 1, file_valid = 0;
+                    unsigned char * audio_data;
+                    netmd_wireformat wireformat;
+                    unsigned char discformat;
+
+                    /* read source */
+                    stat(argv[2], &stat_buf);
+                    puts("reading file: stat o.k.");
```
thp: The debug output here and below should probably be made conditional with some macro?

netmdcli/netmdcli.c
```diff
@@ -410,138 +477,192 @@ int main(int argc, char* argv[])
...
+                    data_size = (size_t)stat_buf.st_size;
+                    printf("file size : %d bytes\n", data_size);
+                    data = (unsigned char *)malloc(data_size);
+                    if(data == NULL) puts("error allocating memory for file input");
```
thp: Would need to clean up here and/or break in case we can't allocate memory, right now this just continues below with the invalid pointer (`fread(data, data_size, 1, f);`).

netmdcli/netmdcli.c
```diff
@@ -410,138 +477,192 @@ int main(int argc, char* argv[])
...
+                    if(!f) puts("cannot open file");
+                    else puts("file opened");
+                    fread(data, data_size, 1, f);
+                    puts("file read");
+                    fclose(f);
+                    puts("file closed");
+
+                    file_valid = audio_supported(data, &wireformat, &discformat, &need_conversion);
+                    if(file_valid == 1) puts("supported audio file detected");
+                    else puts("audio file unknown or not supported");
```
thp: Again, if it's unsupported, we probably need to break out and clean up and not continue.

thp (changes requested):
Still needs some work, but a nice feature to be added.

libnetmd/secure.c
tharp: On my tests , some time ago, i just played around with the tiimeout for a "normal length" track of 4 - 5 minutes. My current test wav file has 16MB (1:31 min) and needs round about 27 seconds for packet transfer. I think a safe value could be 20 sec/minute (pcm stereo) or 2 sec / 1MB packet size. I´ve tested witk my RH1, i don´t know if other/older devices are slower.

Afaik the largest minidiscs are 80 min., i didn´t transfer a 80 min track yet. Maybe it is worth a test with 1600 sec timeout (20 sec/min).

netmdcli/netmdcli.c
tharp: I just used these functions 1:1 as we use it in libhimd/qhimdtransfer, but you´re right, it´s easier to read.

netmdcli/netmdcli.c
tharp: O.K., good point. Possibly a wave file could have more chunks as just the standard "riff" and "data" chunks. Maybe we can use a combination of both. Every chunk stores its chunksize, so we could jump to the next chunk and then check if it´s the data chunk.

netmdcli/netmdcli.c
tharp: O.K., good point.

netmdcli/netmdcli.c
tharp: In addition to that, i think we are also missing the framesize for mono pcm in netmd_get_framesize(). Mono transfers should not work in the  current version, but, as i said, i´ve not tested mono.

netmdcli/netmdcli.c
tharp: Maybe we could use netmd_log() function with level NETMD_LOG_DEBUG.

netmdcli/netmdcli.c
tharp: Yes, you´re right, i just expanded the existing "send" routine but forgot to add some "break on error" parts. This is also missing in the current version.

glaubitz:
@tharp Please reformat and rewrite your commit messages. The first line should contain a short summary without a full stop and no longer than ~65 columns. Use the present tense and proper capitalization.

Additional description may be added in a block paragraph starting from the third line, the second line remains blank.

This is the usual formatting style for commit messages used by most open source projects.

tharp:
can be closed without merging, see pull request #28

thp:
Closing as per the latest comment, let's continue the discussion in #28 (note that you could also force-push the rebased commits onto your pull request branch to have the pull request updated, but whatever works for you).

PR #28, opened 2016-12-25
Netmd download support in netmdcli

tharp:
Same as pull request #27, but as i had to rebase to edit commit messages i decided to put my changes in  2 new commits, one for libnetmd related stuff and the other for the changes in netmdcli.

The usb timeout value is set to ca. 3 1/3 sec. per 1MB (exactly 3 1/3 sec per 1.000.000 bytes).
My test pcm file has ca. 16MB and needs 46 - 48 seconds to transfer using my RH1 on usb 2.0 port in WinXP in a vm. Usb timeout is set to 53 sec. for this file.
Detailed debug messages (including data size, usb timeout and real time consumption) can be displayed using -v oprtion.


libnetmd/secure.c
```diff
@@ -513,6 +523,10 @@ netmd_error netmd_secure_send_track(netmd_dev_handle *dev,
     totalbytes = netmd_get_frame_size(wireformat) * frames + packet_count * 24U;
     netmd_copy_doubleword_to_buffer(&buf, totalbytes, 0);
 
+    /* log time consumption for debugging/analysing usb timeout value */
+    time(&starttime);
```
thp: It's just for logging, but it'd be better to use monotonic clocks, so that this doesn't fail in case the system time is adjusted via NTP or through other means during the transfer.

See also: https://www.softwariness.com/articles/monotonic-clocks-windows-and-posix/
We could use GLib's function, but libnetmd doesn't really require glib (which I just looked at for this comment):
https://developer.gnome.org/glib/unstable/glib-Date-and-Time-Functions.html#g-get-monotonic-time

thp (commented):
Code-wise, apart from the monotonic clock (which we might not need, since it's just for logging?) the code looks good.

I have not tested if this actually works with a device, though -- maybe we should have a small LP2 and LP4 file in the repo's test data (with CC-licensed audio) that we can use for netmd uploads here?

tharp:
Tested it with my RH1 in a virtual WinXP.
Again, i used the clock/time only for finding a good usb timeout value and left it in because it may help for debugging/testing.
 
Afair there is a LP2.wav file for testing. I cannot find it yet, but i think it is somewere on the linux-minidisk wiki on the fu-berlin site.

tharp:
O.K., I found a bug.
If padding is needed/performed in netmd_prepare_packets(...) gcry_cipher_encrypt(...) call will run out of scope and libgcrypt will crash with a segmentation fault.
But this could also happen in the current version when padding to multiple of 8 bytes is needed.

For now, i just increased buffer size to prevent this, but maybe we should rethink about calculating the transfer data size based on the number of frames.

Also, on my tests some transfers failed. I don't know why but sometimes libusb_bulk_transfer(...) call in netmd_transfer_song_packets(...) fails with error code -1 (LIBUSB_ERROR_IO).
All other commands are fine.
So i changed the debug message log level to NETMD_LOG_VERBOSE.

P.S.: the commit of the win32 binary file was a mistake, i'm not very familiar with github.
I just wanted to test a "binary file release" but didn't realize that it was added and commited to my branch

tharp:
I successfully transferred large pcm wav files to the MZ-RH1 in multiple packets.

Short summary what i changed and why:
The MZ-RH1 seems to not support file transfer in multiple packets, but libusb will fail with input/output error on very large packets.
The solution for this is to build and transfer the packets as they were just one packet, but splitted into several parts.
Libusb splits the bulk transfer data into smaller packets (called urbs) of 16384 bytes.
Every urb except the last one must be completely filled with data, so the chunksize of the netmd packets must be multiple of 16384 bytes.
The header (24 bytes for length, iv and key) must be present in the first packet only.
Audio data must be continuously encrypted from the first to the last packet using the same key.
Frames must not be truncated if incomplete.

In summary when playing with chunksize:
The transfer should succeed if all of the following contitions met
- no truncated frames (padding if needed)
- chunksize multiple of 16384 (incl. 24 byte header in first packet)
- total transfer size = framesize * frames + 24 (header)
- total transfer size passed to the device in the command for the download request must match the size really transferred through libusb_bulk_transfer() function
- continuously audio data encryption with same key

Thomas

P.S.: Successfully tested with a 184Mb wav file (round about 17:30 minutes audio) on debian to my MZ-RH1.
Transfer ratio is round about 2:1 in relation to real time recording (8:45 mins for my file)

h3x4d3c1m4l:
I tried several Python scripts from multiple sources, but using none of them I could successfully write a PCM wav file to my MZ-N510.

This seems to work! No errors and my MDS-JE530 plays the songs I wrote to the MiniDisc perfectly fine. Hopefully this will be merged some day.

Please do let me know if I can be of any help to have this merged :)

h3x4d3c1m4l:
I made myself a script that automates sending a couple of wav's. It seems to crash after a few songs...
EDIT: after having ejected the disc and trying to check what it actually has written, the TOC seemed corrupt (C14 TOC error appears on my player).

https://pastebin.com/myRZfQ3E

Replugging the device fixes the issue for a while. Could there be some sort of timeout that might be too low for my device?

EDIT2: I changed the LIBUSB timeout from 1000 ms to 10000 ms and it already succesfully sent 10 songs to the NetMD...

glaubitz:
@thp @karcherm I think we should get this PR from @tharp reviewed and merged so that the changes by @vuori from #62 can be merged on top.

glaubitz:
@karcherm @thp It seems like @tharp isn't active at the moment, so we should review the PR without him and fix potential problems later.

glaubitz:
I'm working on cleaning Thomas Arp's patches now. This will take some time as it seems some changes that a logically independent were accidentally squashed into one commit.

glaubitz:
Cleaned up patches by Thomas Arp have been merged now, closing this PR.

PR #29, opened 2016-12-25
Remove glib dependency for libnetmd/netmdcli/qhimdtransfer

thp:
We don't link against those libraries, maybe we don't need them there.


thp:
Since this builds in CI and seems straightforward enough, I'll merge this one right away, hope that is okay :)

glaubitz:
Yes, this one's okay.

PR #30, opened 2017-01-24
Fix Travic CI dependencies for osx builds

tharp:
Travis CI throws out an error for osx builds because of missing qt4 libraries but qt5 is available in the build environment


tharp:
O.K., doesn't work because of missing qmake.
Seems to be something wrong/missing in the Travis CI build environment.

Can be closed or merged if we want to use qt5 for osx builds.

thp:
+1, let's go with Qt 5. I've tested locally with Qt 5 on macOS 10.12 and it works nicely (in fact, Qt 4 has been removed and is unmaintained AFAIK). The build on Travis has been restarted now, let's see if it passes.

thp:
Merged #31.

PR #33, opened 2017-02-01
Detect minidisc devices on linux

tharp:
I added an autodetection routine for linux.
It is based on sending/recieving DBus messages.

At application start: enumerate installed /dev/sd* drives and check if they are usb connected, vendor is "Sony" and Model is "Hi-MD"
Usb vendor/product id's cannot be read this way but i dont want to use further libraries like libudev.

While application runtime: use udev rules and a notify script to send dbus message to a running qhimdtransfer instance if a minidisc deice is added or removed.
I provided the rules file and notify script in qhimdtransfer/linux/etc

Also i changed the devicelist_changed() slot in qhimdmainwindow.cpp to not open any autodetected device.
As i tested detection routine is very fast. On netmd devices the TOC has not been read completely by the device, tracklist will appear corrupt.
Himd devices are not mounted because udev didn't finish processing. Mountpoint autodetection will fail.

All autodetected devices are displayed in the staus bar.
User has to wait for the device to be ready (netmd: TOC is loaded, himd: device is mounted).
Maybe we should add a messagebox to inform the user to wait before trying to open.

Thomas

### DIFF
--- a/qhimdtransfer/linux/etc/minidisc_notify.sh
+++ b/qhimdtransfer/linux/etc/minidisc_notify.sh
@@ -1,0 +1,25 @@
+#!/bin/sh -e
+
+ case $ACTION in
+         add)
+                 ACT=AddMDDevice
+                 ;;
+         remove)
+                 ACT=RemoveMDDevice
+                 ;;
+         *)
+                 ;;
+ esac
+
+# run as current user, else dbus message cannot be recieved
+username=`who | head -n 1 | cut -d " " -f 1`
+export HOME=/home/$username
+export XAUTHORITY=$HOME/.Xauthority
+export DISPLAY=:0.0
+
+# convert vendor and product id to int
+vid=$(echo $((0x$ID_VENDOR_ID)))
+pid=$(echo $((0x$ID_MODEL_ID)))
+
+# send message to session bus, do not wait for a reply !!!
+su $username -c "qdbus --session org.linux-minidisc.qhimdtransfer /QHiMDUnixDetection $ACT $DEVNAME $vid $pid &" -m
--- a/qhimdtransfer/qhimdunixdetection.cpp
+++ b/qhimdtransfer/qhimdunixdetection.cpp
@@ -1,0 +1,426 @@
+#include <QtDBus/QtDBus>
+#include <qhimddetection.h>
+#include <QVariantList>
+#include <QDebug>
+#include <QApplication>
+
+/* define some constants for UDisks2 bdus operations */
+#define UDISK_SERVICE "org.freedesktop.UDisks2"
+#define UDISK_PATH "/org/freedesktop/UDisks2"
+#define UDISK_INTERFACE "org.freedesktop.UDisks2"
+#define UDISK_PROPERTIES "org.freedesktop.DBus.Properties"
+#define UDISK_INTERFACE_FILESYSTEM "org.freedesktop.UDisks2.Filesystem"
+#define UDISK_INTERFACE_BLOCK "org.freedesktop.UDisks2.Block"
+#define UDISK_INTERFACE_DRIVE "org.freedesktop.UDisks2.Drive"
+#define UDISK_DEVICE_PATH "/org/freedesktop/UDisks2/block_devices/"
+
+
+/* qhimdtransfer adaptor class, can be accessed through dbus session bus with
+ * service=org.linux-minidisc.qhimdtransfer
+ * path=/QHiMDUnixDetection
+ * interface=org.linux-minidisc.qhimdtransfer.QHiMDUnixDetection
+ *
+ * following methods provided
+ * "AddMDDevice" and "RemoveMDDevice"
+ * with args: QString deviceFile, int vid, int pid
+ */
+class QHiMDAdaptor: public QDBusAbstractAdaptor
+{
+    Q_OBJECT
+    Q_CLASSINFO("D-Bus Interface", "org.linux-minidisc.qhimdtransfer")
+    Q_CLASSINFO("D-Bus Introspection", ""
+"  <interface name=\"org.linux-minidisc.qhimdtransfer\">\n"
+"    <method name=\"AddMDDevice\">\n"
+"      <arg direction=\"in\" type=\"s\" name=\"deviceFile\"/>\n"
+"      <arg direction=\"in\" type=\"i\" name=\"vendorId\"/>\n"
+"      <arg direction=\"in\" type=\"i\" name=\"productId\"/>\n"
+"    </method>\n"
+"    <method name=\"RemoveMDDevice\">\n"
+"      <arg direction=\"in\" type=\"s\" name=\"deviceFile\"/>\n"
+"      <arg direction=\"in\" type=\"i\" name=\"vendorId\"/>\n"
+"      <arg direction=\"in\" type=\"i\" name=\"productId\"/>\n"
+"    </method>\n"
+"  </interface>\n"
+        "")
+public:
+    QHiMDAdaptor(QObject *parent) : QDBusAbstractAdaptor(parent) {}
+    virtual ~QHiMDAdaptor() {}
+};
+
+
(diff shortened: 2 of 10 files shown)


thp:
Wouldn't it be better to make it so that QHiMDTransfer just watches (using udev monitor) for device connect/disconnect events? https://www.freedesktop.org/software/systemd/man/udev_monitor_new_from_netlink.html#

The current solution makes many assumptions (i'll add separate comments), including that only one user is logged in. If you really wanted to go that route, at least use logind's D-Bus API to find which user is "active":
https://www.freedesktop.org/wiki/Software/systemd/logind/

Also, the name `qhimdunixdetection.cpp` is misleading, it should probably be called `qhimdlinuxdetection.cpp` (given that macOS is Unix, and the *BSDs are also Unix, but none of them will work with this implementation, so it's really a Linux-only solution, which is fine).

My proposal would be that udev doesn't trigger any shell script or anything, but QHiMDTransfer uses the udev-monitor library from systemd to react to hotplug events, this way, no overhead and no shell scripts would be used in case it's not running and a device is hotplugged, but still work correctly if QHiMDTransfer is running (or even if multiple instances, maybe for different users, are running).

At the bottom of this page is a nice tutorial for how to use udev for montoring hotplug events (it's written for hidraw devices, but this could easily be customized to work for the MD use case):

http://www.signal11.us/oss/udev/

qhimdtransfer/linux/etc/minidisc_notify.sh
```diff
@@ -0,0 +1,25 @@
...
+         remove)
+                 ACT=RemoveMDDevice
+                 ;;
+         *)
+                 ;;
+ esac
+
+# run as current user, else dbus message cannot be recieved
+username=`who | head -n 1 | cut -d " " -f 1`
+export HOME=/home/$username
```
thp, line 16: It's not as easy as that.

Simple example: Imagine that somebody (for whatever reason) is logged into an X session using the "root" account (it's usually `/root/`, not `/home/root/`, but even `/root/` is just convention).

Better solutions:
http://superuser.com/questions/484277/get-home-directory-by-username
http://unix.stackexchange.com/questions/247576/how-to-get-home-given-user

qhimdtransfer/linux/etc/minidisc_notify.sh
```diff
@@ -0,0 +1,25 @@
...
+                 ;;
+         remove)
+                 ACT=RemoveMDDevice
+                 ;;
+         *)
+                 ;;
+ esac
+
+# run as current user, else dbus message cannot be recieved
+username=`who | head -n 1 | cut -d " " -f 1`
```
thp, line 15: As mentioned in another comment, this makes too many assumptions, and doesn't work if there are multiple users logged in (either locally via session switching, locally on different VTs or even SSH sessions, etc..).

qhimdtransfer/linux/etc/minidisc_notify.sh
```diff
@@ -0,0 +1,25 @@
...
+export HOME=/home/$username
+export XAUTHORITY=$HOME/.Xauthority
+export DISPLAY=:0.0
+
+# convert vendor and product id to int
+vid=$(echo $((0x$ID_VENDOR_ID)))
```
thp, line 21: The `$(echo ...)` is redundant, this also seems to work for me (at least in `bash`):

    vid=$((0x$ID_VENDOR_ID))

That said, you could just leave it as string and do the conversion on the C++ side.

qhimdtransfer/qhimdunixdetection.cpp
```diff
@@ -0,0 +1,426 @@
+#include <QtDBus/QtDBus>
+#include <qhimddetection.h>
+#include <QVariantList>
+#include <QDebug>
+#include <QApplication>
+
+/* define some constants for UDisks2 bdus operations */
```
thp, line 7: Maybe instead of hardcoding those here and using D-Bus directly, maybe use this library?

https://udisks.freedesktop.org/docs/latest/UDisksClient.html

Alternatively, this might or might not be useful:

https://github.com/udisks-qt/udisks-qt

tharp:
> Wouldn't it be better to make it so that QHiMDTransfer just watches (using udev monitor) for device connect/disconnect events? https://www.freedesktop.org/software/systemd/man/udev_monitor_new_from_netlink.html#

udev_monitor does not throw events/signals, we need to implement our own polling method to get any changes. Just "watching" for connect/disconnect events does not work this way.
I don't like that polling stuff.
By analysing dbus objects i found UDisks2 ObjectManager sending signals "InterfaceAdded" and "InterfaceRemoved" we can use for himd devices, but this will not work for netmd devices as they are not block devices/disks.

> The current solution makes many assumptions (i'll add separate comments), including that only one user is logged in. If you really wanted to go that route, at least use logind's D-Bus API to find which user is "active":
https://www.freedesktop.org/wiki/Software/systemd/logind/

ok, finding the "active" user is a good point.

> Also, the name qhimdunixdetection.cpp is misleading, it should probably be called qhimdlinuxdetection.cpp (given that macOS is Unix, and the *BSDs are also Unix, but none of them will work with this implementation, so it's really a Linux-only solution, which is fine).

ok, i didn't think about other os are unix but not linux.

> My proposal would be that udev doesn't trigger any shell script or anything, but QHiMDTransfer uses the udev-monitor library from systemd to react to hotplug events, this way, no overhead and no shell scripts would be used in case it's not running and a device is hotplugged, but still work correctly if QHiMDTransfer is running (or even if multiple instances, maybe for different users, are running).

> At the bottom of this page is a nice tutorial for how to use udev for montoring hotplug events (it's written for hidraw devices, but this could easily be customized to work for the MD use case):

> http://www.signal11.us/oss/udev/

If we want to use a polling object we can go this way. 
Another way is to use the hotplug implementation of libusb, which can be used similar to the udev monitor implementation by polling for changes.
Using libusb is mostly platform independent, but libusb hotplug events are not supported on all platforms. If mac is supported we can use it for mac and linux, windows detection works fine as it is. 
I don't know if there is a way to get device driver /dev/sd* according to the libusb_device for himd devices, but as stated above i found UDisks2 dbus signals we can use for himd devices on linux.
Or maybe some of the libusb interface descriptors can be used to get the device driver from USisks2 via dbus message.

thp:
You can use [QSocketNotifier](http://doc.qt.io/qt-5/qsocketnotifier.html) with `QSocketNotifier::Read ` to watch the file descriptor from the monitor (`udev_monitor_get_fd()`) in an efficient fashion and get Qt-friendly signals, and handle the event (`udev_monitor_receive_device()`) in the Qt signal handler.

I didn't know about libusb hotplug events, but it sounds like a good solution (even better if it also works on macOS and Windows). And even if the `libusb_device` returned from libusb's hotplug API doesn't lead to a device/mountpoint name, we can use those events to trigger a re-scan of existing drives, or use UDisks2 as you suggested for finding the mountpoint.

tharp:
> I didn't know about libusb hotplug events, but it sounds like a good solution (even better if it also works on macOS and Windows). And even if the libusb_device returned from libusb's hotplug API doesn't lead to a device/mountpoint name, we can use those events to trigger a re-scan of existing drives, or use UDisks2 as you suggested for finding the mountpoint.

libusb hotplug events are supported since version 1.0.16 but i didn't test it yet.
I don't have a mac, can anyone please check if it is available on mac by calling
libusb_has_capability(LIBUSB_CAP_HAS_HOTPLUG) in a test application?
At least for netmd devices the autodetection code could be completely platform independent. 

[http://libusb.sourceforge.net/api-1.0/hotplug.html](url)

Edit: I took a look at the libusb source code. It seems that hotplug events are implemented for linux, mac and windows but capability may depend on the os version.
I think i'll give it a try with libusb.

thp:
Looking at https://github.com/libusb/libusb/blob/master/libusb/os/darwin_usb.c, there is indeed an implementation for macOS using IOKit.

tharp:
Now another try using libusb hotplug event support in #34

PR #34, opened 2017-02-09
Detect all supported minidisc devices platform independent

tharp:
In relation to the conversation in #33, 
now another try of autodetection using libusb hotplug event support.

Should work on linux, mac and windows (tested on linux only, yet).
No need of platform dependent code for netmd devices.
For himd devices  platform dependent add/remove routines needed to find the corresponding path (device file, driver file, handle, ...or what else needed.).
Including himd detection code for linux using dbus messages and updated code for windows.

Edit: Forgot to change add_himddevice()/remove_himd_device() calls in win and mac detection code.
I'll fix this soon.


tharp:
I also rearranged netmd device enumeration using libusb hotplug feature instead of netmd_init().
This way we do not need to reenumerate (remove all and enumerate new) netmd devices.
They can be added and removed the same way as himd devices.

Current changes will skip mac autodetection of adding devices by path.
They will be detected and added by libusb device and name but without a path.
User will be asked for a path at opening.
Disconnection will also be detected.
I don't know if there is a way to get a path from a libusb_device on mac.

PR #37, opened 2017-03-30
netmdcli: check arguments before out-of-bounds access

rxrbln:
currently netmdcli simply access argv arguments before checking
that they were actually supplied by the user, leading to segfaults

### DIFF
--- a/netmdcli/netmdcli.c
+++ b/netmdcli/netmdcli.c
@@ -1,4 +1,5 @@
-/* netmd.c
+/* netmdcli.c
+ *      Copyright (C) 2017 René Rebe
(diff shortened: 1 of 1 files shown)


karcherm:
Please apply the comments I made on the other pull request as long as they are still applicable.

thp:
..or maybe switch the code to using getopt, getopt_long or argp?
The only problem is probably MSVC compatibility, but maybe there's a drop-in library for that?

https://www.gnu.org/software/libc/manual/html_node/Parsing-Program-Arguments.html

thp:
Merged in #42.

PR #36, opened 2017-03-28
really scan for the highest numbered HiMD index file

rxrbln:
currently the code apparently only takes the first random index as returned by the OS, however, we need the highest number'ed index file.

PR #41, opened 2018-10-23
libhimd: Scan for highest-numbered track index file

thp:
Cleaned up version of #36.


thp:
(note that if it's easier to review and merge, I can also do separate PRs, but i thought I'd just put those all together here)

Issue #47, opened 2019-01-11
Sony NetMD MZ-NE410: Song does not transfer properly, apparently


IngwiePhoenix:
So first, I tried this:

```
Ingwie@Ingwies-Macbook-Pro.local ~/W/G/linux-minidisc $ ./netmdcli/netmdcli send /Applications/Ygopro/sound/deck.mp3 
Net MD Walkman
Raw title: test 
Disc Title: test

No error
No error
No error
No error
No error
No error
No error
netmd_exch_message: netmd_poll failed
Unknown Error
netmd_exch_message: netmd_poll failed
Unknown Error
No error
No error
```

Then I unplugged the device and scoped through the songs, and only found three songs... Well, turns out that is what the program reports - as well as a crash?

```
Ingwie@Ingwies-Macbook-Pro.local ~/W/G/linux-minidisc $ ./netmdcli/netmdcli 
Net MD Walkman
Raw title: test 
Disc Title: test

Track  0: TrPROT    LP4 - 01:40:02 - Intro
Track  1: TrPROT    LP4 - 04:12:30 - Algemeine Musik - Raucherlied
Track  2: TrPROT    LP4 - 03:51:31 - Dirk Krempel - Kakalakenlidchen2

--Empty Groups--


fish: './netmdcli/netmdcli' terminated by signal SIGABRT (Abort)
```

Running LLDB over it:
```
Ingwie@Ingwies-Macbook-Pro.local ~/W/G/linux-minidisc $ lldb ./netmdcli/netmdcli
(lldb) target create "./netmdcli/netmdcli"
Current executable set to './netmdcli/netmdcli' (x86_64).
(lldb) run
Process 31503 launched: './netmdcli/netmdcli' (x86_64)
Net MD Walkman
Raw title: test 
Disc Title: test

Track  0: TrPROT    LP4 - 01:40:02 - Intro
Track  1: TrPROT    LP4 - 04:12:30 - Algemeine Musik - Raucherlied
Track  2: TrPROT    LP4 - 03:51:31 - Dirk Krempel - Kakalakenlidchen2

--Empty Groups--


Process 31503 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff7c09523e libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
[...]

PR #48, opened 2019-01-12
Fix out-of-bounds access in netmd_request_title() (Contributes to #47)

thp:
QHiMDTransfer crashed when trying to access a NetMD device (see #47):

In `QNetMDDevice::open()`, there is a buffer allocated on the stack:

    char buffer[256];

This buffer is used to request title information to find out the
number of tracks on the NetMD device:

    while(netmd_request_title(devh, i, buffer, sizeof(buffer)) >= 0)

Then, in `libnetmd/trackinformation.c`, in `netmd_request_title()`, at
the end of the function, the byte after this buffer is touched:

    buffer[size] = 0;

Work around this by cleaning up the limit calculation a bit and only
writing up to `size-1` bytes from the title to the buffer (so there is
always a trailing `'\0'`).

In case of #47, this was running on macOS, so it probably had enough stack protection built in to detect that issue (that probably existed for a long time).

### DIFF
--- a/libnetmd/trackinformation.c
+++ b/libnetmd/trackinformation.c
@@ -113,15 +113,18 @@ int netmd_request_title(netmd_dev_handle* dev, const uint16_t track, char* buffe
     if(title_size == 0 || title_size == 0x13)
         return -1; /* bail early somethings wrong or no track */
 
-    if(title_size > size)
+    int title_response_header_size = 25;
+    const char *title_text = title + title_response_header_size;
+    size_t required_size = title_size - title_response_header_size;
+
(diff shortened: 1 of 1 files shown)


IngwiePhoenix:
Ouch! This is way too classic: Using `size-1` for the meantime should work. :)

IngwiePhoenix:
Also regarding stack - and in general, memory - protection: Apparently Darwin (especially it's BSD part) does a very solid job of memory management and faulty memory detection. It does have some other flaws (I am always down to 2GB of free memory on 16GB...?) but it is definitively solid on this kind of issue.

PR #49, opened 2019-01-12
Add GUI MP3 downloading support, move code to libhimd

thp:
This moves the `himd_writemp3()` function (and its dependencies) from `himdcli` to `libhimd`, so it can be used in QHiMDTransfer for downloading MP3 files to a HiMD device.

It's as simple as it can get (UI freezes while transfer is in progress, no visual feedback), but should be a good starting point for improving this in the future. Right now, NetMD support is not implemented, but this could be added in the future (probably needs moving the `"send"` command's code from `netmdcli` to `libnetmd` in a similar fashion).


IngwiePhoenix:
Moving the code to the library would mean that other tools could utilize it too. Im considering making an app with FLTK to make copying mp3s from audiobooks to a HiMD easier, which is what I use MiniDiscs mostly for.

IngwiePhoenix:
Awesome!

I can't seem to delete files, but I can probably delete them via the player itself. Thank you for merging this to upstream! Tested and verified on Mac OS.

PR #50, opened 2019-02-16
Add `pkg-config` dependency on macOS

ticky:
Mac builds require `pkg-config` in order to locate other dependencies, and this is _usually_ installed if you’ve done any other software development, but it’s better to be explicit about it here! 😃


thp (approved):
Thanks!

PR #51, opened 2019-04-08
Fixes to Italian translation

martinwguy:
Moe idiomatic and courteous, (only bank machines and automatic phones say "attendere prego" :) and a few spellos


thp (approved):
I don't know any Italian, but presumed good :+1:

PR #56, opened 2020-03-20
libnetmd: Handle encoding and decoding of track and disc titles

ticky:
Wow! We’ve come a long, long way from when I first wrote this! I’ve converted this to a draft PR so it’s clear there’s ongoing work and discussion here.

Ongoing discussion relates to implementing handling for group titles, which are [tricky due to how they’re stored on the MiniDisc](https://github.com/linux-minidisc/linux-minidisc/pull/56#issuecomment-609107256). My [current proposal is that the routines for parsing disc information be updated to handle both title fields at once](https://github.com/linux-minidisc/linux-minidisc/pull/56#issuecomment-629058430), so that there’s no possibility of losing data on any given group edit action.

---

Similarly to #55 (and tested on the same discs!), this updates `libnetmd` to handle conversion of track and disc titles from UTF-8 to their target encodings in a similar way to `libhimd`.

From extensive research over the course of several weeks, I have determined* that narrow MiniDisc titles are JIS X0201 encoded. JIS X0201 is an older Japanese text encoding which includes all of the basic ASCII set for latin text, as well as a set of Japanese Katakana characters.

MiniDiscs also feature a “wide” title field, which is encoded as Shift JIS for greater character support (including Hiragana and many Kanji) supported primarily by high-end LCD remote controls (RM-MC35ELK and the like, the “K” in the model number seems to be the important bit!).

Previously, `libnetmd` has only handled narrow titles, and has put whatever bytes you give it into the player’s narrow title field unadulterated. This means that, if you have a modern system configured for UTF-8, you can set titles which the player has no means of displaying!

This pull request updates `libnetmd` behind the scenes such that the `netmd_set_title` and `netmd_set_disc_title` will encode your UTF-8 inputs as either JIS X0201, or Shift JIS, depending on their contents, and then write the result to the field they fit in. It additionally updates the `netmd_request_title` and `request_disc_title` functions to decode the JIS encoded text into UTF-8, and automatically select the correct field to draw on.

For the character set conversion, this adds a dependency on `glib` to `libnetmd`, but given `libhimd`’s reliance this doesn’t seem like a big deal.

Finally, to improve the ergonomics of handling Japanese text on a modern computer, these functions will also convert the half-width Katakana characters into full-width Unicode Katakana for display, and full-width Unicode Katakana into half-width Katakana when writing half-width titles.

Prior to this pull request, `netmdcli` has this to show for a Japanese-titled disc on a modern UTF-8 system:
image

With this pull request applied, `netmdcli` now shows this for the same disc:
image

---

*I am certainly not the first to discover this fact - [Wikipedia](https://en.wikipedia.org/w/index.php?title=Katakana&oldid=948283472#Half-width_kana) notes the encoding of half-width Katakana, and [win nmd, for instance](http://winnmd.net/winnmd.htm) makes reference to the Shift JIS field’s existence. I have verified this by testing many strings against my Shift JIS incapable MZ-M10 player, and with my Shift JIS capable RM-MC35ELK LCD remote, as well as by observing and testing the behaviour of both English and Japanese versions of SonicStage with Japanese text. Nobody on the internet seemingly outright states that the encodings are these two encodings, but the character set support and behaviour of both players, and SonicStage, make me certain.

### DIFF
--- a/libnetmd/kataconv.c
+++ b/libnetmd/kataconv.c
@@ -204,7 +204,7 @@ kata_full_to_half(uint8_t* input)
     convert(input, FULL_HEAD, FULL_TAIL, HALF_HEAD, HALF_TAIL);
 }
 
-// rewrites full-width katakana to half-width in place
+// rewrites half-width katakana to full-width in place
 void
 kata_half_to_full(uint8_t* input)
 {
--- a/libnetmd/kataconv.c
+++ b/libnetmd/kataconv.c
@@ -21,7 +21,7 @@
  *
  */
 
-#include <stdint.h>
+#include "kataconv.h"
 
 struct tail {
     uint8_t a;
--- a/libnetmd/kataconv.c
+++ b/libnetmd/kataconv.c
@@ -1,0 +1,212 @@
+/*
+ * kataconv.c
+ *
+ * This file is part of libnetmd, a library for accessing Sony NetMD devices.
+ *
+ * Copyright (C) 2020 Charlie Somerville
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ */
+
+#include <stdint.h>
+
+struct tail {
+    uint8_t a;
+    uint8_t b;
+};
+
+static const uint8_t FULL_HEAD = 0xe3;
+
+// second and third bytes of each full-width katakana
+// the first byte is always 0xe3, so we don't need to include it in this list
+static const struct tail FULL_TAIL[] = {
+    { 0x80, 0x81 }, // "、"
+    { 0x80, 0x82 }, // "。"
+    { 0x80, 0x8c }, // "「"
+    { 0x80, 0x8d }, // "」"
+    { 0x82, 0x9b }, // "゛"
+    { 0x82, 0x9c }, // "゜"
+    { 0x82, 0xa1 }, // "ァ"
+    { 0x82, 0xa2 }, // "ア"
+    { 0x82, 0xa3 }, // "ィ"
+    { 0x82, 0xa4 }, // "イ"
+    { 0x82, 0xa5 }, // "ゥ"
+    { 0x82, 0xa6 }, // "ウ"
+    { 0x82, 0xa7 }, // "ェ"
+    { 0x82, 0xa8 }, // "エ"
+    { 0x82, 0xa9 }, // "ォ"
+    { 0x82, 0xaa }, // "オ"
+    { 0x82, 0xab }, // "カ"
+    { 0x82, 0xad }, // "キ"
+    { 0x82, 0xaf }, // "ク"
+    { 0x82, 0xb1 }, // "ケ"
+    { 0x82, 0xb3 }, // "コ"
+    { 0x82, 0xb5 }, // "サ"
+    { 0x82, 0xb7 }, // "シ"
+    { 0x82, 0xb9 }, // "ス"
+    { 0x82, 0xbb }, // "セ"
+    { 0x82, 0xbd }, // "ソ"
+    { 0x82, 0xbf }, // "タ"
+    { 0x83, 0x81 }, // "チ"
+    { 0x83, 0x83 }, // "ッ"
+    { 0x83, 0x84 }, // "ツ"
+    { 0x83, 0x86 }, // "テ"
+    { 0x83, 0x88 }, // "ト"
+    { 0x83, 0x8a }, // "ナ"
+    { 0x83, 0x8b }, // "ニ"
+    { 0x83, 0x8c }, // "ヌ"
+    { 0x83, 0x8d }, // "ネ"
+    { 0x83, 0x8e }, // "ノ"
+    { 0x83, 0x8f }, // "ハ"
+    { 0x83, 0x92 }, // "ヒ"
+    { 0x83, 0x95 }, // "フ"
+    { 0x83, 0x98 }, // "ヘ"
+    { 0x83, 0x9b }, // "ホ"
+    { 0x83, 0x9e }, // "マ"
+    { 0x83, 0x9f }, // "ミ"
+    { 0x83, 0xa0 }, // "ム"
+    { 0x83, 0xa1 }, // "メ"
+    { 0x83, 0xa2 }, // "モ"
+    { 0x83, 0xa3 }, // "ャ"
+    { 0x83, 0xa4 }, // "ヤ"
+    { 0x83, 0xa5 }, // "ュ"
+    { 0x83, 0xa6 }, // "ユ"
+    { 0x83, 0xa7 }, // "ョ"
+    { 0x83, 0xa8 }, // "ヨ"
+    { 0x83, 0xa9 }, // "ラ"
+    { 0x83, 0xaa }, // "リ"
+    { 0x83, 0xab }, // "ル"
+    { 0x83, 0xac }, // "レ"
+    { 0x83, 0xad }, // "ロ"
+    { 0x83, 0xaf }, // "ワ"
+    { 0x83, 0xb2 }, // "ヲ"
+    { 0x83, 0xb3 }, // "ン"
+    { 0x83, 0xbb }, // "・"
+    { 0x83, 0xbc }, // "ー"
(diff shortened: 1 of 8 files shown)


thp:
Would it be easy to also fix writing disc titles? Decode from the system locale's encoding to Unicode and then encode into Shift-JIS, dropping/replacing any characters that are non-ASCII and non-katakana characters?

Is there anywhere a spec that says that all MiniDisc titles use Shift-JIS or are there regional differences? (i.e. do we need a switch?)

For German-language titles, ß, ü, ö, ä, Ä, Ü, Ö would be the special characters, but I don't remember if MD equipment supports that (and those can be transliterated to sz, ue, oe, ae, Ae, Ue, Oe if need be).

ticky:
> Would it be easy to also fix writing disc titles?

Writing titles should be doable too! I ran out of time last night and wanted to see how this approach was received. ☺️

> Is there anywhere a spec that says that all MiniDisc titles use Shift-JIS or are there regional differences? (i.e. do we need a switch?)

I’ve not got access to any comprehensive reference material on this, however, on testing with several discs sourced from Japan with Japanese text, as well as a couple of discs titled on American-delivered equipment, I have yet to encounter any which don’t fit into Shift JIS (makes sense given that the first half of single-byte Shift JIS matches ASCII!). It’s possible that there’s an encoding switch in the title header, as there is on Hi-MD, but I can’t find anyone’s notes on the header format, and my gut would be willing to believe that back in 1992 Sony hadn’t thought that through. 😅

I don’t have a good source for European discs with likely-different encoding, and I don’t have a working copy of SonicStage handy to see what it does. I’ll see if I can get that set up, maybe after I look into title writing. Once the infrastructure is there adding a switch should be trivial enough if necessary!

One thing to note is that the Hi-MD library presumes your system is using UTF-8 for I/O, which seems reasonable to me, and a pattern I’ve thus far followed - would leaving the assumption there be reasonable? I don’t think I have a good way to test other system locales!

ticky:
I have confirmed through some testing on my Hi-MD capable player that all these Minidiscs I own are interpreted as being JIS X0201, an encoding which forms the basis for Shift JIS, but does not include the full-width katakana forms; full-width katakana forms are not displayed, and encoding them with Shift JIS doesn’t render correctly on either the player (an American-delivered MZ-M10) or the remote control (an aftermarket RM-MC35ELK). I will update this PR to use JIS X0201 specifically, so that full-width forms give a proper encoding error, rather than discs filled with garbled text. I will also investigate whether conversion of the widths is possible.

libnetmd/kataconv.c
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// rewrites full-width katakana to half-width in place
+void
+kata_full_to_half(uint8_t* input)
+{
+    convert(input, FULL_HEAD, FULL_TAIL, HALF_HEAD, HALF_TAIL);
+}
+
+// rewrites full-width katakana to half-width in place
```
ticky: This comment is the wrong way around, oops!

glaubitz:
Very cool patch set, thanks for this. I'm not sure I have regular MDs with Japanese tracks as I transferred all my music to HiMD, including Japanese tracks.

I have enough MiniDisc hardware and Japanese CD for testing though but I'm not sure I can test this myself in a timely manner.

So, I'll just review the changes and merge them over the weekend.

ticky:
This isn’t quite ready to go, sorry! I’m hoping to have it ready by the end of this weekend, pacific time, but I haven’t had enough focus on it yet.

ticky:
Oh, and I will be able to provide some testing instructions which will compare the patched and unpatched versions, and most likely work across most players. Stay tuned! ☺️

glaubitz:
Okay, great. Thank you. I do have a large number of Japanese CDs and I happen to speak enough Japanese to use SonicStage in Japanese and I even have my old Japanese Windows XP VM in a backup. Although I cannot say how long it takes to get everything set up again :).

ticky:
Very close! I’ve just implemented wide title field support for reading disc and track titles, and for writing track titles. Disc title writing is still to come, but I’ve got to take a break to cook some dinner! 😄

---
Basically the logic is: since the single-byte JIS X0201 encoding appears to be the baseline encoding standard for all MiniDisc players, it has the widest support on devices. Shift JIS (a variable-width encoding) appears to extend the format in a non-mandatory way. Devices without support for wide titles encoded in Shift JIS will just show no title for Shift JIS encoded titles. Shift JIS support seems to be most commonly available on the high-end `RM-MC*K` remotes. For that reason, my code prefers setting JIS X0201 encoded titles where possible, and falls back to Shift JIS where necessary. Shift JIS allows encoding much more of the Japanese writing system, including Hiragana and many Kanji, as well as some Russian and Greek characters, none of which are included in JIS X0201.

By updating this to use UTF-8 for input and output, it now does what most people likely thought it did already, but with much greater support for handling Japanese writing! My hope is that by implementing this here, this improved support will be able to spread to derivative and otherwise related projects like platinum-md and web minidisc!

ticky:
Some late-breaking discovery on my part: it’s actually _definitely_ possible for a disc or track to have both wide and narrow titles simultaneously - I now have a track which is titled "neko" on my M10's display and 「ねこ」 on my RM-MC35ELK. This has some major implications for the shape of this functionality, which will now require some reconsideration!

ticky:
I’ve acquired a Japanese-language copy of SonicStage and confirmed it features a wide/narrow toggle in the UI, and titles in both encodings are possible to add on the same track at the same time. It does not blank or replace wide titles when editing narrow, nor vice versa.

In the interests of simplicity and “doing the right thing” I still believe that letting existing implementations work without modification is the best path, but I will look at extending the API with methods for specifically targeting each field.

glaubitz:
@ticky Thanks for your diligent work on this. Much appreciated!

ticky:
One thing I haven't accounted for yet in this pull request is group names - I guessed that this would run through the `netmd_set_disc_title` method, and thus work roughly correctly without intervention. Having since looked into it, this assumption was not correct; it has its own implementation, `netmd_write_disc_header`,  which sends a slightly different request to the player.

Now, I could potentially apply a similar approach to my fixes for `netmd_set_disc_title` and `netmd_set_title`, but given it’s possible for the disc title and individual group titles to have different encodings, this poses a bit of a problem - _any single one_ of these titles requiring the wide field would require me to write _all_ of them to the wide field, and more importantly, in my testing, **the narrow title field is always treated as the canonical version** as far as track indexes go. In cases where a wide field disagrees, it is ignored. If a wide field is missing, it is ignored. If a wide field is set but no narrow field is, it is treated as the title of the entire disc, without splitting up groups.

This means that, either way, for this to support the wide field for groups, it always needs to be able to write to the narrow field at the same time, even if only so that track index information is present. Thus, this step of the process is a much larger and much more complex one - requiring a change to a data structure which persists throughout the library and applications, and requiring changes to the logic of writing the header information to keep the two versions synchronised.

For this reason, I suspect without a sudden blast of inspiration that I’m not going to polish this one off this weekend. I will persist, however, in my efforts. Thank you for your patience while I work all this out - I don’t mean for it to drag on, only for it to be as correct as I can make it! 😄

thp:
Good job so far :)

ticky:
Finally looking at this again! Thank you for your patience, I really want to make this happen.

---

One question I’d love input on from the maintainers is this:

I’ve attempted to make the changes to the CLI portion not break anything, i.e. the behaviour should be the same as before, but not require any specific new implementations to work nicely.

In the `libnetmd` API, I’ve added methods for directly setting wide or narrow titles, but the signatures of the existing methods have not been changed; you can still call `netmd_set_title` with a device handle, a track number, and a buffer, and it’ll still do what `netmd_set_title` used to do, but better.

As for setting groups, I think the best approach to make some minor API changes; the methods for parsing disc titles will now have a need to have both wide and narrow titles parsed, and there’s not really any good that can come from allowing these to be separate functions, and potentially allow getting out of sync.

For instance, `netmd_parse_disc_title`, `netmd_parse_group` and `netmd_parse_trackinformation` would be candidates for this; these are the functions responsible for accepting the existing title string, parsing out the group information, and hydrating the `minidisc` struct, with corresponding `netmd_group` array. Now, these functions are exposed in the `libnetmd` header, so changing their signature gives me pause - `libnetmd` doesn’t appear to produce a library a third party is obviously using directly, but the possibility is there. As far as _internal_ usages go, nothing published reaches this far into the guts; `netmdcli` and `QNetMDDevice` only call in as far as `netmd_initialize_disc_info`, and that API in particular has no need to change.

So, would there be any issues with a breaking change (to consumers of the API doing what I would call unusual-sounding things, but still a hypothetically-public API nonetheless!) on those methods, adding support for the wide title field? My goal would be to have them run over both titles in a single pass, populating that `netmd_group` with both titles, and update `netmd_initialize_disc_info` to handle that transparently.

It’d look something vaguely like this:

```diff
-void netmd_parse_disc_title(minidisc* md, char* title, size_t title_length);
+void netmd_parse_disc_title(minidisc* md, char* title, char* wide_title, size_t title_length);

-void netmd_parse_group(minidisc* md, char* group, int* group_count);
+void netmd_parse_group(minidisc* md, char* group, char* wide_group, int* group_count);

-void netmd_parse_trackinformation(minidisc* md, char* group_name, int* group_count, char* tracks);
+void netmd_parse_trackinformation(minidisc* md, char* group_name, char* wide_group_name, int* group_count, char* tracks);
```

There would be some other changes, too, but these are the low-hanging fruit which look like they’ll determine the shape of the rest of the work.

PR #58, opened 2020-03-23
QNetMDTracksModel: Avoid crash due to permission error

thp:
This fixes a crash when the current user doesn't have permission to open the NetMD device.

Steps to reproduce:

1. Remove the `/etc/udev/rules.d/netmd.rules` file or remove the user from the `plugdev` group.
2. Run QHiMDTransfer

### DIFF
--- a/qhimdtransfer/qmdmodel.cpp
+++ b/qhimdtransfer/qmdmodel.cpp
@@ -129,8 +129,10 @@ QString QNetMDTracksModel::open(QMDDevice * device)
         ret = ndev->open();
     }
 
-    if(!ret.isEmpty())
+    if(!ret.isEmpty()) {
         close();
+        return ret;
+    }
 
     /* fetch track info for all tracks first, getting track info inside data() function is very slow */
     for(; i < ndev->trackCount(); i++)


Invictaz:
@gavinbenda can you merge these fixes in your Platinum-MD ?

gavinbenda:
> @gavinbenda can you merge these fixes in your Platinum-MD ?

I really only use netmdcli in Platinum-MD - I don't think this code is included anywhere, as it's part of qhimdtransfer in linux-minidisc.

Invictaz:
@gavinbenda Understood. I see finally some new developments in the minidisc code so I was wondering if you are keeping track of it.

glaubitz:
@gavinbenda Plantinum-MD looks great. Glad that our code is useful for other projects, too.

If you have any improvements for the linux-minidisc code, feel free to open PRs.

I have asked @vuori to merge upstream his improvements so that we can reduce merge all the improvements people made back upstream. I'm very happy to see all these improvments!

glaubitz (commented):
@thp Is there a reason why we don't immediately want to return from ```QNetMDTracksModel``` when ```device``` is NULL?

Invictaz:
@glaubitz Can you help Gavin port downloackhack.py into Platinum-MD please? He does not have an MZ-RH1 to test. I do have one but lack all the skills you guys have. I'm still using the @thp testing version that has upload functionality.

glaubitz:
@Invictaz The Python code is not really my terrain. It was supposed to be a proof-of-concept only back when it was written. The functionality was later ported to libnetmd, albeit some features in libnetmd need more fixing, but @vuori has fixed many issues.

Invictaz:
@glaubitz Thanks for the answer. To my knowledge that downloadhack.py was not touched in ages. 
Users e-mail me because they do not listen to music anymore on the minidisc, but use the RH1 for archival purposes. With the Minidisc becoming older a valid archival method is necessary. 
SonicStage 6.0 was only Japanese and I failed to install it, even after modifying some ini files.

thp:
@Invictaz Please file a separate issue for the downloadhack.py-to-libnetmd porting.

thp:
@glaubitz Updated PR.

1. `device` and `ndev` are actually non-`NULL` (wrong assumption on my part)
2. `ret = ndev->open();` returns a non-empty string on permission error
3.  `if(!ret.isEmpty()) close();` causes `ndev` to become `NULL`

So the fix for this particular issue was to `return ret;` when we `close()` due to an error message returned from `ndev->open()`.

glaubitz:
Great. Looks good to me. Shall I merge or do you want to merge it yourself?

vuori:
re: downloadhack, I haven't touched that and I don't think it's really useful except as reference. It should probably be removed or moved to some "old stuff" directory at some point so people don't try to use it.

Invictaz:
@vuori it's the best we have to get songs of the device. If I could be of any assistance (without lending or destroying my MZ-RH1) let me know.

PR #60, opened 2020-03-23
Port NetMD scripts to Python 3

thp:
WIP port. Mostly using `2to3`, plus fixed up so that `lsmd.py` works.

### DIFF
--- a/netmd/libnetmd.py
+++ b/netmd/libnetmd.py
@@ -1,5 +1,5 @@
 import libusb1
-from cStringIO import StringIO
+from io import StringIO
 from time import sleep
 from struct import pack
 try:
@@ -12,7 +12,7 @@ import array
 import random
 
 def dump(data):
-    if isinstance(data, basestring):
+    if isinstance(data, str):
         result = ' '.join(['%02x' % (ord(x), ) for x in data])
     else:
         result = repr(data)
@@ -20,8 +20,8 @@ def dump(data):
 
 class defaultUploadEvents:
     def progress(self, current):
-        print 'Done: %x/%x (%.02f%%)' % (current, self.total,
-                                         current/float(self.total) * 100)
+        print('Done: %x/%x (%.02f%%)' % (current, self.total,
+                                         current/float(self.total) * 100))
     def trackinfo(self, frames, bytes, format):
         self.total = bytes;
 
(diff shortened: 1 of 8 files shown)


thp:
***bump*** any chance of getting this reviewed/merged?

glaubitz:
Can you retitle the commit 35d3d3369c2bf6add6b608a87e30cbdb29b66d11?

It's currently called "Merge branch 'master' into python3".

Also, it looks like that the commit changes more than it should.

glaubitz:
Thanks for the clean-up! Looks good to me now.

Unless @karcherm has anything else to add or complain, we can merge this now!

netmd/libnetmd.py
```diff
@@ -12,16 +12,16 @@
 import random
 
 def dump(data):
-    if isinstance(data, basestring):
+    if isinstance(data, str):
```
karcherm, line 15: This should surely be `if isinstance(data, bytes)`, we don't use (unicode) strings here.

netmd/libusb1.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, member_dict):
         reverse_dict = {}
```
karcherm, line 15: I don't think we should 2to3 an ancient version of https://github.com/vpelletier/python-libusb1, which is also available via pypi, but just depend on `libusb1`.

netmd/usb1.py
```diff
@@ -23,7 +23,7 @@
 from ctypes import byref, create_string_buffer, c_int, sizeof, POINTER, \
     create_unicode_buffer, c_wchar, cast, c_uint16, c_ubyte, string_at, \
     addressof, c_void_p, cdll
-from cStringIO import StringIO
+from io import StringIO
```
karcherm, line 26: The comment about just depending on a recent libusb1 also applies to this file.

netmd/lsmd.py
```diff
@@ -38,36 +38,36 @@ def timeToFrames(time_tuple):
...
             codec, channel_count = md_iface.getTrackEncoding(real_track)
             flags = md_iface.getTrackFlags(real_track)
-            print '%s%03i: %02i:%02i:%02i+%03i %s %s %s %s %s' % (prefix,
+            print('%s%03i: %02i:%02i:%02i+%03i %s %s %s %s %s' % (prefix,
                 track, hour, minute, second, sample, codec_name_dict[codec],
                 channel_count_dict[channel_count], flag_dict[flags],
-                md_iface.getTrackTitle(real_track).decode('shift_jis_2004'),
-                md_iface.getTrackTitle(real_track, True).decode('shift_jis_2004'))
+                md_iface.getTrackTitle(real_track),
+                md_iface.getTrackTitle(real_track, True)))
```
karcherm, line 67: getTrackTitle scans using `%x`, which returns a `bytes` object. Why don't we `decode` it anymore. Isn't that adding ugly `b'...'` to the output and print unreadable hex escapes for titles with Japanese chars?

karcherm (changes requested):

PR #62, opened 2020-03-29
NetMD download support fixes (on top of #28)

vuori:
As discussed with @glaubitz in e-mail, this is the most important stuff from my fork. This is based on Thomas Arp's PR https://github.com/linux-minidisc/linux-minidisc/pull/28 . It's probably a good idea to merge that first to get a nicer diff.

There are two important fixes here:

  1. The protocol uses 3DES in CBC mode, but the block chaining was not being done correctly, leading to noise at transfer block boundaries.
  2. Padding at end of data was incorrect, which led to a burst of noise at end of track.

Additionally there's some code simplication and comments, and I reduced the transfer block size to 1MB to improve transfer reliability. It's still nowhere near perfect though: libusb will error out after 5-10 transfers and leave the player's USB interface hung.

There's quite a bit of work that could be done with libnetmd and netmdcli wrt reliability and such, but getting this 95% working download support merged at this point would make further PRs a lot easier to put together.

edit: the transfer chunk size idea comes from https://github.com/linux-minidisc/linux-minidisc/pull/19 so I think this PR will supercede that.


ticky:
Super excited to see this! Happy to test these fixes with my equipment if that’s helpful. 😄

libnetmd/secure.c
```diff
@@ -439,26 +480,39 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
...
         }
 
-        /* generate key */
-        gcry_randomize(rand, sizeof(rand), GCRY_STRONG_RANDOM);
-        gcry_cipher_decrypt(key_handle, next->key, 8, rand, sizeof(rand));
-
         /* crypt data */
         memcpy(next->iv, iv, 8);
+        memcpy(next->key, key, 8);
         gcry_cipher_setiv(data_handle, iv, 8);
```
karcherm, line 486: If we move this line before the loop, we don't need to keep track of the IV over chunks ourselfs, so line 501 `"memcpy(iv, next->data + chunksize -8, 8)"` can be removed.

libnetmd/secure.c
```diff
@@ -349,77 +349,118 @@ size_t netmd_get_frame_size(netmd_wireformat wireformat)
...
     int error;
     int transferred = 0;
+    int first_packet = 1;
 
     p = packets;
     while (p != NULL) {
         /* length + key + iv + data */
-        packet_size = 8 + 8 + 8 + p->length;
+        if(first_packet)                                     // length, key and iv in first packet only
+            packet_size = 8 + 8 + 8 + p->length;
```
karcherm, line 366: If we encrypt the whole track with one key and IV, it would seem wise to adjust the data model, so that netmd_track_packets does no longer contain "per-packet" key and iv. We should pass them separately as a track property, just like `full_length` has also been added as parameter to this function.

libnetmd/secure.c
```diff
@@ -349,77 +349,118 @@ size_t netmd_get_frame_size(netmd_wireformat wireformat)
...
 
         /* build packet... */
-        netmd_copy_quadword_to_buffer(&buf, p->length);
-        memcpy(buf, p->key, 8);
-        memcpy(buf + 8, p->iv, 8);
-        memcpy(buf + 16, p->data, p->length);
+        if(first_packet) {                                   // lenght, key and iv in first packet only
+            netmd_copy_quadword_to_buffer(&buf, full_length);
+            memcpy(buf, p->key, 8);
+            memcpy(buf + 8, p->iv, 8);
```
karcherm, line 377: Consider advancing buf by 16 after this line, so that the data memcpy does not need to be different for the first and subsequent packets.

netmdcli/netmdcli.c
```diff
@@ -168,6 +171,80 @@ void retailmac(unsigned char *rootkey, unsigned char *hostnonce,
...
+    if(strncmp("RIFF", file, 4) != 0 || strncmp("WAVE", file+8, 4) != 0 || strncmp("fmt ", file+12, 4) != 0)
+        return 0;                                             /* no valid WAV file or fmt chunk missing*/
+
+    if(leword16(file+20) == 1)                                /* PCM */
+    {
+        *conversion = 1;                                      /* needs conversion (byte swapping) for pcm raw data from wav file*/
+        *wireformat = NETMD_WIREFORMAT_PCM;
+        if(leword32(file+24) != 44100)                        /* sample rate not 44k1*/
+            return 0;
+        if(leword16(file+34) != 16)                           /* bitrate not 16bit */
```
karcherm, line 210:
```suggestion
        if(leword16(file+34) != 16)                           /* bit depth not 16bit */
```

netmdcli/netmdcli.c
```diff
@@ -410,138 +490,216 @@ int main(int argc, char* argv[])
...
+                                netmd_log(NETMD_LOG_ERROR, "cannot open audio file\n" );
+                                file_valid = 0;
+                                free(data);
+                            }
+                        }
+                    }
+
+                    /* copy file to buffer */
+                    if(file_valid) {
+                        memset(data, 0, data_size+8);
```
karcherm, line 561: Why +8 here, when we have +2048 in line 546?

netmdcli/netmdcli.c
```diff
@@ -410,138 +490,216 @@ int main(int argc, char* argv[])
...
-            puts(netmd_strerror(error));
-
-            /* forget key */
-            error = netmd_secure_session_key_forget(devh);
-            puts(netmd_strerror(error));
-
-            /* leave session */
-            error = netmd_secure_leave_session(devh);
-            puts(netmd_strerror(error));
+                    netmd_ekb ekb;
```
karcherm, line 493: The indentation of this block does not match the original file indentation. Maybe (I'm just looking at it in the github web interface) the replacement block here is indented with tabs instead of spaces, and it just looks the same at a tab size of 4. Please don't change the kind of indentation in a file just for a part of it.

karcherm (changes requested):

vuori:
I didn't look at the secure.c suggestions in detail yet, but AFAICT all changes to netmdcli.c are actually part of #28. I can fix the indentation and comments here if needed, but can't really say about the WAV offsets. This parser effectively works for the most commonly encountered WAV file structure, but IMO should be replaced by a more robust one (or, preferably, an audio file handling library) soonish. Hence I never spent much time trying to understand what it's doing…

karcherm:
> all changes to netmdcli.c are actually part of #28.

Yeah, I am sorry for the communication issue caused by just using the web review feature. I talked with cbmuser, and suggested to him to rewrite #28 to not include the temporary addition of the ZIP file, and apply all comments I made to this issue but actually are on changes in #28 to #28 before merging that one. After that, we can rebase your changes on the new master.

I did not intend to push the burden of cleaning up #28 onto you.

glaubitz:
I'm still trying to reach @tharp so he can make the changes that @karcherm requested.

Otherwise, we will have to figure out how we can get this PR merged without Thomas Arp.

Sorry for the delay.

glaubitz:
Update: I have now heard back from @tharp and he says, he is currently very busy with other matters and doesn't have much time for software development.

He has asked me to pick up his commits, clean them up and commit them. I will do that over the weekend.

Then we can merge this PR.

netmdcli/netmdcli.c
glaubitz: It seems that the code uses the designation "bitrate" a lot. So it might be a better idea to rename it to bit_depth in a later commit.

netmdcli/netmdcli.c
glaubitz: I assume that should be +2048 to clear the whole buffer?

thp:
There's now also https://github.com/vuori/linux-minidisc/pull/8 in the downstream repo.

I know everybody is likely super busy and/or has other things do to, but maybe we should just skip some style issues and merge (and cleanup style issues separately) so that things don't diverge too much? (there's now @vuori's repo, which some seem to start using and @cybercase ported libnetmd to typescript: https://github.com/cybercase/netmd-js)

glaubitz:
I have cleaned most things up now. There is just one commit which messes up the indentation left and when I try to clean it up, git becomes confused which is why I stopped.

https://github.com/glaubitz/linux-minidisc/tree/netmd-download-support-rebased

I will try to clean up the rest later today and then open the PR.

glaubitz:
Rebased patch set submitted as #87.

Issue #64, opened 2020-04-11
Please make a recent Windows build


Invictaz:
There is no recent Windows build. There is no Travis build bot for it. 
Can you please fix this? 

https://travis-ci.org/github/glaubitz/linux-minidisc

thp:
There is some Travis CI setup, not sure why it doesn't build in this repo. I pushed the tags from this repo here to https://github.com/thp/linux-minidisc/releases (will take a while to build), but eventually I need to look into why it doesn't build/upload into this repo.

Possibly the API Key needs to be updated in `.travis.yml`.

thp:
Note that since this project was moved into an organization, the Travis CI page should be:

https://travis-ci.org/github/linux-minidisc/linux-minidisc

However, this is not set up yet, let's see if I have permissions to do that.

Invictaz:
Great, I will check later on your progress

Invictaz:
@glaubitz any success?

glaubitz:
@Invictaz Could you please stop doing this?

I'm working on this in my free time and I have a day job as well plus a private life. I have started working on it, but I haven't finished it as the code needs more clean-up. When it's finished, you will be automatically notified.

Please be patient. Thank you.

deenine:
I have taken a look at this, the travis build doesnt build anymore as far as I can tell, so I have put up a trusty64 vm to test (https://github.com/deenine/lmd-vagrant/blob/main/lmd.yml - please ignore the reference to my fork in lmd.yml, the below is discussing the upstream linux-mindisc in this repo).

I think it needs `cmake` and `gettext` adding to the list of packages installed in `install_dependencies.sh`

It then fails due to not being able to find glib-2.0, which is due to the `PKG_CONFIG_PATH` set in `build/cross_wrapper_mingw.sh` not being passed to pkg-config. A quick hack is setting `PKG_CONFIG = PKG_CONFIG_PATH=/opt/mingw64/lib/pkgconfig pkg-config` in `libhimd/libhimd.pro`, `himdcli/himdcli.pro` `netmdcli/netmdcli.pro` and `himdcli/himdcli.pro`, but obviously this isnt a good option as it will break the linux/osx builds.

I'm kinda stuck at this point, due to not being a linux or qmake expert, any ideas how to fix this properly?

thp:
New Windows builds will be possible with https://github.com/linux-minidisc/linux-minidisc/pull/79

[...]

PR #70, opened 2020-12-29
Ensure pkg-config is correctly queried for `libgcrypt` and `libid3tag`

ticky:
Previously, pkg-config was not invoked for `libgcrypt`, and both it and `libid3tag` was not required by all its dependent modules, and therefore non-standard installation locations for these dependencies would fail to be found. This updates the project files to correctly wire through the dependencies.

This patch does the following:

- `build/libgcrypt.pri`: Use `qmake`'s `PKGCONFIG` variable rather than `LIBS`, unless `pkg-config` is unable to find `libgcrypt`, in which case we fall back to using the `libgrypt-config` utility instead, with logic borrowed from `qmake` itself.
- `himcdli/himdcli.pro`: Remove `libid3tag.pri` and `libz`, as these dependencies are used only by `libhimd`
- `libhimd/libhimd.pro`: Add `libid3tag.pri` and `libz` (a dependency of `libid3tag`), as they [are used by libhimd](https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/libhimd/mp3tools.c#L26)
- `netmdcli/netmdcli.pro`: Add `libgcrypt.pri` as it [is used by netmdcli](https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/netmdcli/netmdcli.c#L22)

I've tested this successfully on a macOS machine, but I believe it's acting in a fairly standard manner, and shouldn't change any results on other platforms. On further iteration, this now appears to build correctly on our Linux and Windows cross-compilation CI!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
       compiler: gcc
       env: BUILD_TYPE=linux-cross-mingw32 CONFIGURE_COMMAND="sh build/cross_wrapper_mingw.sh qmake" COMPILE_COMMAND="sh build/cross_wrapper_mingw.sh make"
     - os: osx
-      osx_image: xcode10
+      osx_image: xcode10.2
       language: cpp
       compiler: clang
       env: BUILD_TYPE=osx-native-clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
       compiler: gcc
       env: BUILD_TYPE=linux-cross-mingw32 CONFIGURE_COMMAND="sh build/cross_wrapper_mingw.sh qmake" COMPILE_COMMAND="sh build/cross_wrapper_mingw.sh make"
     - os: osx
-      osx_image: xcode7.2
+      osx_image: xcode10
       language: cpp
       compiler: clang
       env: BUILD_TYPE=osx-native-clang
--- a/build/libgcrypt.pri
+++ b/build/libgcrypt.pri
@@ -1,5 +1,5 @@
 !without_gcrypt {
-  LIBS += -lgcrypt
+  PKGCONFIG += libgcrypt
   DEFINES += CONFIG_WITH_GCRYPT
 }
 else: !build_pass: message(You disabled gcrypt: No PCM and ATRAC transfer will be supported)
--- a/build/libgcrypt.pri
+++ b/build/libgcrypt.pri
@@ -1,5 +1,28 @@
 !without_gcrypt {
-  PKGCONFIG += libgcrypt
+  # We use qt's own check to see if pkg-config can find libgcrypt, and if it can't, fall back to libgcrypt-config
+  # https://github.com/qt/qtbase/blob/5.3/mkspecs/features/qt_functions.prf#L323-L336
+  packagesExist(libgcrypt) {
+    PKGCONFIG += libgcrypt
+  } else {
+    # The following is based upon the qt5 link_pkgconfig mkspec logic:
+    # https://github.com/qt/qtbase/blob/5.3/mkspecs/features/link_pkgconfig.prf#L12-L27
+    GCRYPT_CONFIG_CFLAGS = $$system(libgcrypt-config --cflags)
+
+    GCRYPT_CONFIG_INCLUDEPATH = $$find(GCRYPT_CONFIG_CFLAGS, ^-I.*)
+    GCRYPT_CONFIG_INCLUDEPATH ~= s/^-I(.*)/\\1/g
+
+    GCRYPT_CONFIG_DEFINES = $$find(GCRYPT_CONFIG_CFLAGS, ^-D.*)
+    GCRYPT_CONFIG_DEFINES ~= s/^-D(.*)/\\1/g
+
+    GCRYPT_CONFIG_CFLAGS ~= s/^-[ID].*//g
+
+    INCLUDEPATH *= $$GCRYPT_CONFIG_INCLUDEPATH
+    DEFINES *= $$GCRYPT_CONFIG_DEFINES
+
+    QMAKE_CXXFLAGS += $$GCRYPT_CONFIG_CFLAGS
+    QMAKE_CFLAGS += $$GCRYPT_CONFIG_CFLAGS
+    LIBS += $$system(libgcrypt-config --libs)
+  }
   DEFINES += CONFIG_WITH_GCRYPT
 }
 else: !build_pass: message(You disabled gcrypt: No PCM and ATRAC transfer will be supported)
--- a/build/libz.pri
+++ b/build/libz.pri
@@ -1,5 +1,12 @@
 unix:!mac {
     PKGCONFIG += zlib
 } else {
-    LIBS += -lz
+    # Windows builds with QT4 will get confused by the link order of zlib,
+    # so we use QMAKE_LFLAGS instead to force it to be dead last.
+    # https://stackoverflow.com/a/31483107
+    !greaterThan(QT_MAJOR_VERSION, 4):!unix {
+        QMAKE_LFLAGS += -lz
+    } else {
+        LIBS += -lz
+    }
 }
--- a/build/libz.pri
+++ b/build/libz.pri
@@ -1,12 +1,5 @@
 unix:!mac {
     PKGCONFIG += zlib
 } else {
-    # Windows builds with QT4 will get confused by the link order of zlib,
-    # so we use QMAKE_LFLAGS instead to force it to be dead last.
-    # https://stackoverflow.com/a/31483107
-    !greaterThan(QT_MAJOR_VERSION, 4):!unix {
-        QMAKE_LFLAGS += -lz
-    } else {
-        LIBS += -lz
-    }
+    LIBS += -lz
 }
--- a/himdcli/himdcli.pro
+++ b/himdcli/himdcli.pro
@@ -6,9 +6,7 @@ CONFIG += console link_pkgconfig link_prl
 SOURCES += himdcli.c
 
 include(../libhimd/use_libhimd.pri)
-include(../build/libid3tag.pri)
 include(../build/libmad.pri)
-include(../build/libz.pri)
 include(../build/libglib.pri)
 include(../build/installunix.pri)
 include(../build/common.pri)
--- a/himdcli/himdcli.pro
+++ b/himdcli/himdcli.pro
@@ -6,7 +6,6 @@ CONFIG += console link_pkgconfig link_prl
 SOURCES += himdcli.c
 
 include(../libhimd/use_libhimd.pri)
-include(../build/libid3tag.pri)
 include(../build/libmad.pri)
 include(../build/libz.pri)
 include(../build/libglib.pri)
--- a/himdcli/himdcli.pro
+++ b/himdcli/himdcli.pro
@@ -6,8 +6,9 @@ CONFIG += console link_pkgconfig link_prl
 SOURCES += himdcli.c
 
 include(../libhimd/use_libhimd.pri)
+include(../build/libid3tag.pri)
 include(../build/libmad.pri)
-include(../build/libglib.pri)
 include(../build/libz.pri)
+include(../build/libglib.pri)
 include(../build/installunix.pri)
 include(../build/common.pri)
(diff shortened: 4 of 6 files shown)


thp (approved):
Looks good to me. However, the CI builds fail, which might be unrelated to this PR, but something to keep in mind before merging.

ticky:
Hmm, looks like it _is_ related to locating `libgcrypt` using `pkg-config` on that old version of Ubuntu, is there something special you have to do there?

ticky:
It appears that the maintainer of `libgcrypt` was for a long time a curmudgeon who resisted adding `pkg-config` compatibility, so only very recent versions have it, and even then, it's broken as of the current version without patches from master. If you're lucky enough your distro maintainer might've patched it to do the right thing, but in this case we don't seem so lucky.

I'm not sure I'm familiar enough with qmake's syntax to make this work in both cases, it appears that falling back to using the output of `libgcrypt-config --libs` and `libgcrypt-config --cflags` in place of that of `pkg-config` would do the job, I'm just not sure how you'd wire that into the `PKGCONFIG` stuff qmake does.

ticky:
I believe I got it working; in 966509a027ef2efe20b978e78c6f2abb36d389d6 I've used part of qt's own logic for handling this, so we query whether `pkg-config` knows about `libgcrypt`, and if not, it will fall back to calling for `ligbcrypt-config` using the same sort of logic as qt itself uses for the `pkg-config` metadata when it's present.

This allows us to use the more consistent `pkg-config` behaviour where supported, and fall back in other cases to the older method.

I've tested this locally by making `packagesExist` false, but I'll wait on the pending CI run to call it a complete win :)

ticky:
Well, Linux builds are working, but the cross-compilation for Windows is now broken. Fix building for one OS and two more break, I guess! Argh!

deenine:
I looked at the travis error for the mingw32 build:
```
i686-w64-mingw32-g++ -L/opt/mingw32//lib/ -Wl,-s -Wl,-subsystem,console -o release/himdcli.exe release/himdcli.o  -L'../libhimd/release' -L'../libhimd' -L../libhimd/release -L../libhimd -lhimd -lz -lid3tag -lmad -L/opt/mingw32/lib -lgcrypt -lgpg-error -lglib-2.0 
/opt/mingw32//lib//libid3tag.a(util.o):util.c:(.text+0x1e9): undefined reference to `compress2'
/opt/mingw32//lib//libid3tag.a(util.o):util.c:(.text+0x290): undefined reference to `uncompress'
collect2: error: ld returned 1 exit status
```

This can be fixed by moving `-lz` to the end of the line:
```
travis@vagrant-ubuntu-trusty-64:~/ticky-lmd/himdcli$ i686-w64-mingw32-g++ -L/opt/mingw32//lib/ -Wl,-s -Wl,-subsystem,console -o release/himdcli.exe release/himdcli.o  -L'../libhimd/release' -L'../libhimd' -L../libhimd/release -L../libhimd -lhimd -lz -lid3tag -lmad -lgcrypt -lgpg-error -L/opt/mingw32/lib -lglib-2.0
/opt/mingw32//lib//libid3tag.a(util.o):util.c:(.text+0x1e9): undefined reference to `compress2'
/opt/mingw32//lib//libid3tag.a(util.o):util.c:(.text+0x290): undefined reference to `uncompress'
collect2: error: ld returned 1 exit status

travis@vagrant-ubuntu-trusty-64:~/ticky-lmd/himdcli$ i686-w64-mingw32-g++ -L/opt/mingw32//lib/ -Wl,-s -Wl,-subsystem,console -o release/himdcli.exe release/himdcli.o  -L'../libhimd/release' -L'../libhimd' -L../libhimd/release -L../libhimd -lhimd -lid3tag -lmad -lgcrypt -lgpg-error -L/opt/mingw32/lib -lglib-2.0 -lz

travis@vagrant-ubuntu-trusty-64:~/ticky-lmd/himdcli$ ls release/
himdcli.exe  himdcli.o
```

ticky:
@deenine thanks for the tip! I'll try swapping the order of the includes, and see what CI thinks!

ticky:
Hmm, changing the order doesn't actually change the order? 😰 what is qmake doing

ticky:
Building with qt 5 might(?) fix this, if this commit is anything to go by: https://github.com/qt/qtbase/commit/9d76beee5be0812fdc8bd53c3145de95b45ab6a6, still trying to find other options

deenine:
This seems really wierd, i didnt think the order should make any difference. QT5 brings a different set of problems.

deenine:
@ticky it looks like your latest patch has just removed `-lz`

```x86_64-w64-mingw32-g++ -lz -L/opt/mingw64//lib/ -Wl,-s -Wl,-subsystem,console -o release/himdcli.exe release/himdcli.o  -L'../libhimd/release' -L'../libhimd' -L../libhimd/release -L../libhimd -lhimd -lid3tag -lmad -L/opt/mingw64/lib -lgcrypt -lgpg-error -lglib-2.0 ```

If you want a resonably representative local test environment, i have a vagrant here, just swap deenine/linux-minidisc for ticky/linux-minidisc in lmd.vagrant. https://github.com/deenine/lmd-vagrant The slight gotcha is that qmake cannot find the path, so you have to hack the pkgconfig path in the .pro files, per: https://github.com/linux-minidisc/linux-minidisc/issues/64#issuecomment-744005247

ticky:
Yeah, okay, StackOverflow lied to me, who'd have thought! Rolling that one back.

ticky:
My only hunch is that maybe I disturbed the state of things with my himdcli.pro changes? Did this all build okay on the main branch...?

ticky:
I'm fast running out of ideas :/ none of this makes any sense to me

On master:

`x86_64-w64-mingw32-g++ -L/opt/mingw64//lib/ -Wl,-s -Wl,-subsystem,console -o release/himdcli.exe release/himdcli.o  -L'../libhimd/release' -L'../libhimd' -L../libhimd/release -L../libhimd -lhimd -lid3tag -lz -lmad -lgcrypt -lglib-2.0 `

In this branch:

`x86_64-w64-mingw32-g++ -L/opt/mingw64//lib/ -Wl,-s -Wl,-subsystem,console -o release/himdcli.exe release/himdcli.o  -L'../libhimd/release' -L'../libhimd' -L../libhimd/release -L../libhimd -lhimd -lz -lid3tag -lmad -L/opt/mingw64/lib -lgcrypt -lgpg-error -lglib-2.0 `

ticky:
I _think_ this may be due to the introduction of the `libid3tag` dependency to `libhimd`, which is what I think actually uses `libz`, maybe adding `libz` to `libhimd` and removing it from `himdcli` (I can't find any libz dependencies in the codebase) will do the trick?

thp:
> I _think_ this may be due to the introduction of the `libid3tag` dependency to `libhimd`, which is what I think actually uses `libz`, maybe adding `libz` to `libhimd` and removing it from `himdcli` (I can't find any libz dependencies in the codebase) will do the trick?

Sounds good, please give it a try. Btw, I'm pretty confident if it helps, we could update the Ubuntu image to a newer release if newer package versions have less build issues (e.g. added pkg-config).

ticky:
I think, in part, the issues with the Windows build stem from the fact that `pkg-config` is being skipped for almost all dependencies (Windows is subject to the `else` block of the `unix` guards). The use of `pkg-config` would remove the need for the additional handling of `libz`, as it would be added by the `pkg-config` call for `libid3tag` itself.

I realise that `pkg-config` is not part of a usual native Windows build tool, but we don't appear to supply any other means of building, so would using it for Windows builds, given an apparent dependence on Unix-style mingw compiler toolchains be acceptable?

The other thing I will try is reversing the order again to more closely match what `pkg-config` does on my own system.

ticky:
Ooh, the Windows build has finally turned green! Hooray! 😇

ticky:
Okay, so in one final mess here, it looks like in the 9 months since the last `master` build, macOS 10.13 (which the Travis builds here use, falling back to `xcode9.2` because `xcode7.2` is in turn [no longer supported by Travis](https://docs.travis-ci.com/user/languages/objective-c#supported-xcode-versions)) fell out of support in Homebrew.

I'm going to bump the Travis configuration to `xcode10.2`, which runs macOS 10.14, which is the oldest version still supported by Homebrew, and hopefully then it'll all turn green.

ticky:
(To be clear, the macOS error isn't to do with the changes in this PR, it's just that Homebrew doesn't generally let you pin package versions, or versions of Homebrew itself, and is (frustratingly at times) a constantly moving target)

ticky:
Phew! Glad that works :)

deenine:
Nice one @ticky!

thp:
Thanks for cleaning this up, nice work!

glaubitz:
I'm not fine with this PR. It contains lots cruft with "Reverts" and "Maybe this works?".

And it's not okay that I wasn't asked to review this. I didn't get any notification.

glaubitz (commented):
This should have never been merged.

Why make a change, revert it and then push both commits.

thp:
@glaubitz I created PR#71, feel free to merge that or just force-push an older revision on top of master.

thp:
I'm of course aware that merging #71 will leave the crufty commits in the history.

Feel free to change history by `git rebase -i` and squashing all PR commits to a single one, since I still think the overall change is a good thing to have -- I won't be touching it, so you can change the history the way it works best for you.

If you only want PRs merged that have your approval, please configure the project accordingly:

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/about-protected-branches

deenine:
@glaubitz suggest using squash-merge when merging commits in future to avoid cruft in the commit history. In general this seems like a positive change, master did not build, now it does?

ticky:
> And it's not okay that I wasn't asked to review this. I didn't get any notification.

There's nothing documented anywhere obvious which says who exactly needs to be asked for a review, so I didn't in fact ask for any particular person's review. It sounds like you might simply have notifications turned off for new pull requests?

> Why make a change, revert it and then push both commits.

Because I am reliant on your CI setup for testing fixes for the other platforms I don't have direct access to, and force pushing is in my experience generally considered rude when collaborating with other people on a branch. I'd have been quite happy to squash the history if asked now that it's working, of course

ticky:
In any case, if a force push to master is decided upon I am happy to rebase and squash this, please let me know.

glaubitz:
> Feel free to change history by `git rebase -i` and squashing all PR commits to a single one, since I still think the overall change is a good thing to have -- I won't be touching it, so you can change the history the way it works best for you.

I'm not arguing against the overall change. I'm just really surprised you merged something which has all that cruft in it. Especially after @karcherm rejected the PR by @tharp because of the cruft in there.

glaubitz:
> > And it's not okay that I wasn't asked to review this. I didn't get any notification.
> 
> There's nothing documented anywhere obvious which says who exactly needs to be asked for a review, so I didn't in fact ask for any particular person's review. It sounds like you might simply have notifications turned off for new pull requests?

I have no idea why I didn't get any notification. I was very sure I had everything turned on and I actually thought that Github would do that by default for a project that I created.

> > Why make a change, revert it and then push both commits.
> 
> Because I am reliant on your CI setup for testing fixes for the other platforms I don't have direct access to, and force pushing is in my experience generally considered rude when collaborating with other people on a branch.

No, force-pushing is not considered rude if it's in your own branch. In fact, it's the common thing people do in large projects.

The point is to keep any cruft and experimenting in your own repository and not in the main repo.

> I'd have been quite happy to squash the history if asked now that it's working, of course

OK.

ticky:
I don't want to get into a protracted argument with you @glaubitz, I just want to help facilitate fixing this. Could you clarify what you would like people to do to rectify this situation?

glaubitz:
> I don't want to get into a protracted argument with you @glaubitz, I just want to help facilitate fixing this. Could you clarify what you would like people to do to rectify this situation?

I have resetted the repository now. Please open a new PR.

glaubitz:
Oh, and make sure me and @karcherm are pulled into the review.

I'll check what's wrong with the notification settings to make sure I'm not missing these things in the future.

PR #72, opened 2021-02-12
Ensure pkg-config is correctly queried

### LINKED ISSUE #69: pkg-config is not queried for `libgcrypt`, `libgpg-error` or `libid3tag`, so those in non-standard install locations will fail
I am attempting to build linux-minidisc on an arm64 Mac, and as of https://github.com/Homebrew/homebrew-core/pull/67170, Qt should support a native build on this platform.

That being said, due to the fact that Homebrew on arm64 no longer installs libraries to `/usr/local/lib`, any installation which isn't using `pkg-config` correctly will likely fail to find the requisite libraries.

In the case of linux-minidisc, `libgcrypt`, `libgpg-error` (a dependency of `libgcrypt`) and `libid3tag` appear to not be using `pkg-config` in this situation, which causes them to fail to be found. I've worked around this by manually passing `INCLUDEPATH` and `LIBS` variables to `qmake`, but this isn't a particularly pleasant solution.

https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/build/libgcrypt.pri

https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/build/libid3tag.pri

ticky:
Previously, pkg-config was not invoked for `libgcrypt`, and both it and `libid3tag` was not required by all its dependent modules, and therefore non-standard installation locations for these dependencies would fail to be found. This updates the project files to correctly wire through the dependencies.

This patch does the following:

- `build/libgcrypt.pri`: Use qmake's `PKGCONFIG` variable rather than `LIBS`
- `libhimd/libhimd.pro`: Add `libid3tag.pri` and `libz.pri` as they [are used by libhimd](https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/libhimd/mp3tools.c#L26)
- `himdcli/himdcli.pro`: Remove `libid3tag.pri` and `libz.pri` as they are moved to `libhimd/libhimd.pro`
- `netmdcli/netmdcli.pro`: Add `libgcrypt.pri` as it [is used by netmdcli](https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/netmdcli/netmdcli.c#L22)

This also uses libgcrypt-config utility if pkg-config metadata is missing for it

This should allow us to use the more-correct pkg-config metadata where available, while maintaining backwards compatibility with earlier versions of libgcrypt which don't provide pkg-config metadata.

Finally, this bumps the Travis configuration to to macOS 10.14 (Using the "xcode10.2" Travis image) as the currently specified image has been retired by Travis, causing all Mac CI builds to break.

This was the result of much experimentation in #70, and has now been rebased for final merging. Fixes #69.

I appear to not be allowed to assign reviewers, so, cc: @glaubitz, @karcherm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
       compiler: gcc
       env: BUILD_TYPE=linux-cross-mingw32 CONFIGURE_COMMAND="sh build/cross_wrapper_mingw.sh qmake" COMPILE_COMMAND="sh build/cross_wrapper_mingw.sh make"
     - os: osx
-      osx_image: xcode7.2
+      osx_image: xcode10.2
       language: cpp
       compiler: clang
       env: BUILD_TYPE=osx-native-clang
--- a/build/libgcrypt.pri
+++ b/build/libgcrypt.pri
@@ -1,5 +1,28 @@
 !without_gcrypt {
-  LIBS += -lgcrypt
+  # We use qt's own check to see if pkg-config can find libgcrypt, and if it can't, fall back to libgcrypt-config
+  # https://github.com/qt/qtbase/blob/5.3/mkspecs/features/qt_functions.prf#L323-L336
+  packagesExist(libgcrypt) {
+    PKGCONFIG += libgcrypt
+  } else {
+    # The following is based upon the qt5 link_pkgconfig mkspec logic:
+    # https://github.com/qt/qtbase/blob/5.3/mkspecs/features/link_pkgconfig.prf#L12-L27
+    GCRYPT_CONFIG_CFLAGS = $$system(libgcrypt-config --cflags)
+
+    GCRYPT_CONFIG_INCLUDEPATH = $$find(GCRYPT_CONFIG_CFLAGS, ^-I.*)
+    GCRYPT_CONFIG_INCLUDEPATH ~= s/^-I(.*)/\\1/g
+
+    GCRYPT_CONFIG_DEFINES = $$find(GCRYPT_CONFIG_CFLAGS, ^-D.*)
+    GCRYPT_CONFIG_DEFINES ~= s/^-D(.*)/\\1/g
+
+    GCRYPT_CONFIG_CFLAGS ~= s/^-[ID].*//g
+
+    INCLUDEPATH *= $$GCRYPT_CONFIG_INCLUDEPATH
+    DEFINES *= $$GCRYPT_CONFIG_DEFINES
+
+    QMAKE_CXXFLAGS += $$GCRYPT_CONFIG_CFLAGS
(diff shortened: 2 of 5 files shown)


glaubitz:
Thanks for your patch.

Can you elaborate what kind of platforms you are thinking of for which there is no pkg-config metadata for libgcrypt but just the libgcrypt-config utility?

I understand the reasoning and I definitely think that libgcrypt should pkgconfig as well. But I'm not sure we want to support every possible target configurations users may have.

glaubitz:
Oh, and while I appreciate such a long explanation, I rather not want to have it in the commit message. Putting it in the description here is fine and having a shorter one in the commit message is enough.

ticky:
> Can you elaborate what kind of platforms you are thinking of for which there is no pkg-config metadata for libgcrypt but just the libgcrypt-config utility?

As noted in https://github.com/linux-minidisc/linux-minidisc/pull/70#issuecomment-751985615, this appears to be the case for several Linux variants where the packagers have not provided pkg-config metadata, including the old version of Ubuntu currently used for CI for this project. libgcrypt itself only got pkg-config compatibility in some of the most recent versions, but it seems clear that's the path forward for them.

> Oh, and while I appreciate such a long explanation, I rather not want to have it in the commit message. Putting it in the description here is fine and having a shorter one in the commit message is enough.

I've rebased to move the extraneous detail.

glaubitz:
> As noted in #70 (comment), this appears to be the case for several Linux variants where the packagers have not provided pkg-config metadata, including the old version of Ubuntu currently used for CI for this project. libgcrypt itself only got pkg-config compatibility in some of the most recent versions, but it seems clear that's the path forward for them.

We should rather then bump the Ubuntu version used for the CI rather than adding such a workaround.

ticky:
I do not use Ubuntu day to day, and I don't have anything but this project's CI to test it against. I don't believe that even current versions of Ubuntu have the necessary patches for `pkg-config` to work out of the box for `libgcrypt`. If you'd like to explore that option I'd welcome it.

thp:
IMHO we should just merge this. For platforms that have `libgcrypt` in pkg-config, it will use/prefer that, and it helps for other platforms.

glaubitz:
> On Apr 21, 2022, at 6:45 PM, Thomas Perl ***@***.***> wrote:
> 
> ﻿
> IMHO we should just merge this. For platforms that have libgcrypt in pkg-config, it will use/prefer that, and it helps for other platforms.
> 
I will look at that tomorrow. There was something about this change I wasn’t happy with.

PR #78, opened 2022-04-04
libnetmd: Fix netmdcli play track bug (Fixes #38)

### LINKED ISSUE #38: netmdcli play track bug
netmdcli play track bug

there is a bug on netmdcli play #track, probably affects other commands that use strtoul function too. Whenever you select a track number, it interpretes it as a hex number rather than decimal.

thp:
Tested by @eRadiofan using Sony MZ-N520.
Tested by @brabes using AIWA AM-NX9.

I personally haven't tested this.

See the discussion in #38.

### DIFF
--- a/libnetmd/playercontrol.c
+++ b/libnetmd/playercontrol.c
@@ -103,7 +103,7 @@ netmd_error netmd_set_track(netmd_dev_handle* dev, const uint16_t track)
                                0x00, 0x00, 0x00, 0x00, 0x00};
     unsigned char buf[255];
 
-    proper_to_bcd(track, request + 9, 2);
+    request[10] = track & 0xff;
 
     /* TODO: error checking */
     netmd_exch_message(dev, request, sizeof(request), buf);


glaubitz:
@karcherm Any comments?

karcherm:
This has been tested on two players. The patch switches from BCD encoding the track number to binary encoding. Let's just apply it and wait whether new reports come in that we broke track numbers on other MD recorders. Otherwise, binary encoding is likely the right thing to do. We might want to use a two-byte encoding though, as the original code used 2-byte (4 digit) BCD. This would mean to add this line together with initializing `track[10]`.

```
    request[9] = track >> 8;
```

thp:
According to
https://www.minidisc.org/md_toc.html one byte might be all that’s needed to
fit the track number, so the upper 8 bits of the 16-bit track number
parameter might anyway always be zero. See quoted section below.

======Notes and ObservationsPopular MiniDisc literature talks about a limit
of 255 tracks per disc, but as can be seen, the limit is actually 255
fragments. If a disc has only one fragment per track, then 255 tracks is
also possible.

The UTOC title structure limits MD track titling to a total of 1785
characters (255 cells * 7 chars/cell) and 255 titles. A disc with 255
titled tracks illustrates these limitations; there would be no room on such
a disc for the disc title, and there could be no more than 7 characters in
any title.

======

Michael Karcher ***@***.***> schrieb am Di. 5. Apr. 2022 um
14:29:

> This has been tested on two players. The patch switches from BCD encoding
> the track number to binary encoding. Let's just apply it and wait whether
> new reports come in that we broke track numbers on other MD recorders.
> Otherwise, binary encoding is likely the right thing to do. We might want
> to use a two-byte encoding though, as the original code used 2-byte (4
> digit) BCD. This would mean to add this line together with initializing
> track[10].
>
>     request[9] = track >> 8;
>
>

karcherm:
> According to https://www.minidisc.org/md_toc.html one byte might be all that’s needed to fit the track number, so the upper 8 bits of the 16-bit track number parameter might anyway always be zero. 

Thanks for looking the stuff up. I accepted the pull request.

PR #79, opened 2022-04-08
Migrate from Travis CI to Github Actions

thp:
Travis CI is dead: https://travis-ci.org/github/linux-minidisc/linux-minidisc

> Since June 15th, 2021, the building on [travis-ci.org](http://www.travis-ci.org/) is ceased. Please use [travis-ci.com](http://www.travis-ci.com/) from now on.

This tries to migrate everything 1:1 to Github Actions. The `${{ secrets.GITHUB_TOKEN }}` mentioned in the file is [automatically provided by Github](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) in the build and allows publishing on Github releases (Travis CI had the API key in there, this can be removed once we migrate to Github Actions).


glaubitz:
Good idea. Is there any way to test this?

thp:
> Good idea. Is there any way to test this?

Working on fixing the build as we speak, Github Actions should show all green once it works (not there yet... :).

glaubitz:
OK, thanks! I am planning to resume cleaning up the patches by @tharp and will hopefully create a PR next week so we can finally move forward.

thp:
It took a bit of trial'n'error and the MinGW cross-compilation PPA for Ubuntu was a bit outdated, so I went with Fedora's nice MinGW packages instead (only libusb and libmad missing, which we build from source).

As part of this, CI now also builds against Qt 5, newer libgcrypt, `mingw-bundledlls` got some new system libraries from Windows for its blacklist, MinGW paths were updated, the Qt 5 platform plugin `qwindows.dll` is packaged with the Windows builds, and the `$BUILD_TYPE` variable has been simplified:

- `linux`: Native Linux build with the default compiler (usually `gcc`)
- `mingw32`: Linux cross-compilation for 32-bit Windows using mingw-w64 from Fedora (in Docker)
- `mingw64`: Linux cross-compilation for 64-bit Windows using mingw-w64 from Fedora (in Docker)
- `macos`: Native macOS build with the default compiler (Apple `clang`)

The build results will be automatically published to Github Releases every time a tag is done.

A ZIP file with the build results (packaged up) is available for each (successful) run in the "Artifacts" section of a run, e.g.: 

https://github.com/linux-minidisc/linux-minidisc/actions/runs/2117714496

(those download links will eventually expire, but the Github Releases files will of course be persistent)

glaubitz:
The final changes look great and I agree that's a very good idea. Thanks for doing this!

However, I think it would be better to clean up the commit list a bit as the PR currently contains the trial-and-error history, i.e. first updating the PPA, then switching to Fedora altogether, first disabling win32 builds, then enabling them again etc.

Maybe it would make sense to squash most of the changes into one commit and keep separate commits only where it makes sense.

glaubitz:
Thanks for the clean-up. Looks good to me now.

@karcherm Do you have any more comments or complaints? If not, we should merge this PR.

PR #80, opened 2022-04-21
Added CMT-M373NT to rules to allow it to connect

thp:
Collecting some random patches from the web (this was also at https://github.com/glaubitz/linux-minidisc/pull/3).

### DIFF
--- a/netmd/etc/netmd.rules
+++ b/netmd/etc/netmd.rules
@@ -108,3 +108,6 @@ ATTRS{idVendor}=="054c", ATTRS{idProduct}=="018a", MODE="0664", GROUP="plugdev"
 
 # Sony NetMD CMT-AH10
 ATTRS{idVendor}=="054c", ATTRS{idProduct}=="021d", MODE="0664", GROUP="plugdev"
+
+# Sony NetMD CMT-M333NT/M373NT
+ATTRS{idVendor}=="054c", ATTRS{idProduct}=="00e7", MODE="0664", GROUP="plugdev"


glaubitz:
Looks good to me.

karcherm:
Unfortunately, we have the list of USB device IDs for NetMD devices in multiple places. Please also add the device ID to netmd/libnetmd.py . We already have the ID in libnetmd/netmd_dev.c , so no change required there.

thp:
That's true. There are also some [more](https://github.com/linux-minidisc/linux-minidisc/blob/master/libnetmd/minidisc.usermap) [outdated](https://github.com/linux-minidisc/linux-minidisc/blob/master/qhimdtransfer/qhimddetection.cpp) places that need to be updated when devices are added.

In my friendly fork (which I'll try to get merged upstream eventually) there's a single place to update it (`devicedb/minidisc.devids`), it generates a udev rules file and data for `libusbmd` (which is a small library that allows e.g. `libnetmd` and QHiMDTransfer to share that information from a single place).

It also writes out the database as netmd-js compatible file (not used yet, but once this is merged, maybe we can get the maintainers to use/copy that file) and the Python netmd implementation (since it's kind of nice that it's self-contained, we could use `ctypes` to load `libusbmd`, but that'd require it to be built, etc...).

https://github.com/thp/linux-minidisc/tree/minidisc-ffwd/devicedb

Adding device with that mechanism still involves a diff in multiple places:

https://github.com/thp/linux-minidisc/commit/b1c556246a2f36a0c93b3bbd283e8af568de7df6

But the way this is achieved is simple:

1. Add the entry to `devicedb/minidisc.devids`
2. Run `make -C devicedb`
3. Commit the result

So feel free to close this PR for now, and we'll "fix it properly" with the `devicedb` and `libusbmd` changes.

PR #87, opened 2022-07-14
NetMD download support fixes by Valtteri Vuorikoski

glaubitz:
This pull request contains a rebased and slightly cleaned up version of the NetMD download support fixes by Valtteri Vuorikoski.

### DIFF
--- a/libnetmd/secure.c
+++ b/libnetmd/secure.c
@@ -409,7 +409,7 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
      * Large sizes cause instability in some players especially with ATRAC3 files. */
     size_t chunksize, packet_data_length, first_chunk = 0x00100000U;
     size_t frame_size = netmd_get_frame_size(format);
-    int padding = 0;
+    int padding = 0, frame_padding = 0;
     netmd_track_packets *last = NULL;
     netmd_track_packets *next = NULL;
 
@@ -456,11 +456,12 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
             }
             /* do not truncate if last frame is incomplete, include padding bytes for DES encryption in size calculation */
             if((data_lenght % frame_size) != 0 || padding != 0) {
-                padding = frame_size - (data_lenght % frame_size) - padding;
-                if(padding < 0)
-                    padding += frame_size;
+                frame_padding = frame_size - (data_lenght % frame_size) - padding;
+                if(frame_padding < 0)
+                    frame_padding += frame_size;
             }
-            chunksize = packet_data_length + padding;
+            chunksize = packet_data_length + frame_padding;
+            netmd_log(NETMD_LOG_VERBOSE, "last packet: packet_data_length=%d, padding=%d, frame_padding=%d, chunksize=%d\n", packet_data_length, padding, frame_padding, chunksize);
         }
 
         /* alloc memory */
--- a/libnetmd/secure.c
+++ b/libnetmd/secure.c
@@ -413,9 +413,13 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
 
     gcry_cipher_hd_t key_handle;
     gcry_cipher_hd_t data_handle;
-    unsigned char iv[8] = { 0 };
-    unsigned char rand[8] = { 0 };
-    unsigned char key[8] = { 0 };
+
+    /* We have no use for "security" (= DRM) so just use constant IV.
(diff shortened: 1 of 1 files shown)


glaubitz:
I have verified that the patches do not negatively affect NetMD transfers. Sending a WAV file to my MZ-RH1 worked without any problems.

libnetmd/secure.c
```diff
@@ -399,23 +399,29 @@ void netmd_transfer_song_packets(netmd_dev_handle *dev,
...
-    unsigned char iv[8] = { 0 };
-    unsigned char rand[8] = { 0 };
-    unsigned char key[8] = { 0 };
+
+    /* We have no use for "security" (= DRM) so just use constant IV.
+     * However, the key has to be randomized, because the device apparently checks
+     * during track commit that the same key is not re-used during a single session. */
+    unsigned char iv[8] = { 0, 0, 0, 0, 0, 0, 0 ,0 };
+    unsigned char raw_key[8] = { 0 }; /* data encryption key */
+    unsigned char key[8] = { 0 }; /* data encryption key wrapped with session key */
```
karcherm, line 424: Why do we expand the single 0 to 8 zeroes only in `iv`, but not in `raw_key` and `key`? I consider both styles valid, but why mix them?

libnetmd/secure.c
```diff
@@ -482,10 +484,21 @@ netmd_error netmd_prepare_packets(unsigned char* data, size_t data_lenght,
...
-        gcry_cipher_setkey(data_handle, rand, sizeof(rand));
-        gcry_cipher_encrypt(data_handle, next->data, chunksize, data + position, packet_data_length);
-        /* set last encrypted block as iv for the next packet */
-        memcpy(iv, data + position + packet_data_length - 8, 8);
+
+        if (chunksize > packet_data_length) {
+            /* If last frame is padded, copy plaintext to chunk buffer and encrypt in place.
+             * This avoids calling gcry_cipher_encrypt() with outsize > insize, which leads
+             * to noise at end of track. */
+            memcpy(next->data, data + position, packet_data_length);
```
karcherm, line 492: Don't we need to explicitly zero out the second part of the buffer? We can't rely on uninitialized memory (next->data) to be always zero-initialized.

Do we need the copy and in-place encryption? Wouldn't
```
    gcry_cipher_encrypt(data_handle, next->data, packet_data_length, data + position, packet_data_length);
    memset(next->data, 0, chunksize - packet_data_length);
```
work equally well?

And finally: Wouldn't this work generally, whether we are on the last packet or not, because on non-last packets, memset is called with a zero length? If that's true, we can remove the whole if/else distinction here.

libnetmd/secure.c
glaubitz: You're right, I'll fix that!

glaubitz: Btw, which style do we prefer? Expansion or single digit?

libnetmd/secure.c
glaubitz: Good question, I don't know. We would have to ask @vuori ;).

vuori: `next->data` is already zeroed on line 470. The zeros (silence) that are used to pad the audio data to MD frame size should also be encrypted. If you encrypt just up to `packet_data_length` the player will attempt to decrypt the still-unencrypted zeros, resulting in noise where the last frame was padded.

Also  `packet_data_length` may not be divisible by cipher block size, since the cipher block is 8 bytes but the typical 16bit/2ch PCM "block" will 2x2 bytes.

libnetmd/secure.c
karcherm: Personally, I very slightly prefer the single-zero style, because it takes less space and allows a trailing comment without creating overly long lines. Furthermore, the human brain can't really comprehend the exact number of zeroes. The overall intention of getting more than one zero is clearly visible both by the braces and the `[8]` after the name of the array.

On the other hand, if the people working most on the code are using quality assurance tools (compiler warnings, style-checkers, static analysis tools) that issue a warning over the single-zero pattern, it is definitely better to expand the zereos than to get a nuisance warning that might prevent noticing relevant warnings. Generally deactivating a "number of initializers too low for array" warning also is a bad idea, becasue we *want* to spot code like

```
    int first_five_primes[5] = {2,3,7,11};  // Oops, someone forgot the 5
```

as soon as possible.

So my verdict is: If a relevant tool used by the current core developers requests that expansion and is otherwise unhappy, go along and *consistently* expand it. On the other hand, if there is no technical reason to expand, keep the short-hand form as it was before.

PR #93, opened 2023-09-05
I tried  to make it work with Fedora 38

BTarn:
What I did is
1. Change from libusb.h to libusb-1.0/libusb.h 
2. Comment out the line that include build/libusb.pri

I'm not a programmer, developer, coder, or anything of that sort, so if what I did is funny please feel free to ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ Makefile*
 qtc-gdbmacros
 /.qmake.stash
 /qhimdtransfer/moc_predefs.h
+basictools/himddiskid
--- a/libnetmd/common.h
+++ b/libnetmd/common.h
@@ -1,7 +1,7 @@
 #ifndef LIBNETMD_COMMON_H
 #define LIBNETMD_COMMON_H
 
-#include <libusb.h>
+#include <libusb-1.0/libusb.h>
 
 /**
    Typedef that nearly all netmd_* functions use to identify the USB connection
--- a/libnetmd/libnetmd.h
+++ b/libnetmd/libnetmd.h
@@ -32,7 +32,7 @@
 #include <errno.h>
 #include <stdint.h>
 
-#include <libusb.h>
+#include <libusb-1.0/libusb.h>
 
 #include "const.h"
 #include "error.h"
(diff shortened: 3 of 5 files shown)


thp:
If you install https://packages.fedoraproject.org/pkgs/libusb-compat-0.1/ then the unmodified source should build.

thp:
What was the initial build error that you were getting?

BTarn:
libusb.h is missing.
I can't find any package that provide libusb.h.

From this web, https://fedoraproject.org/wiki/Changes/Rename_libusb_packages_and_deprecated_old_api, they said that they rename package name to libusb1, so I use package libusb1-devel which provides libusb-1.0/libusb.h.

thp:
> libusb.h is missing. I can't find any package that provide libusb.h.
> 
> From this web, https://fedoraproject.org/wiki/Changes/Rename_libusb_packages_and_deprecated_old_api, they said that they rename package name to libusb1, so I use package libusb1-devel which provides libusb-1.0/libusb.h.

But this should be picked up by [`PKGCONFIG += libusb-1.0`](https://github.com/linux-minidisc/linux-minidisc/blob/master/build/libusb.pri#L2).

According to https://packages.fedoraproject.org/pkgs/libusb1/libusb1-devel/fedora-rawhide.html#files, it ships with a pkg-config file, and downloading https://kojipkgs.fedoraproject.org//packages/libusb1/1.0.26/3.fc39/x86_64/libusb1-devel-1.0.26-3.fc39.x86_64.rpm shows that `/usr/lib64/pkgconfig/libusb-1.0.pc` has this content:

```
prefix=/usr
exec_prefix=/usr
libdir=/usr/lib64
includedir=/usr/include

Name: libusb-1.0
Description: C API for USB device access from Linux, Mac OS X, Windows, OpenBSD/NetBSD and Solaris userspace
Version: 1.0.26
Libs: -L${libdir} -lusb-1.0
Libs.private: -ludev 
Cflags: -I${includedir}/libusb-1.0
```

The line `Cflags: -I${includedir}/libusb-1.0` is important here, as it signifies that `libusb-1.0` should be part of the compiler include path, so `libusb.h` should be found via this mechanism, there's no need to specify `libusb-1.0`.

In other words, these changes are wrong when using `pkg-config` (which is the preferred way of finding library dependencies):

```
-#include <libusb.h>
+#include <libusb-1.0/libusb.h>
```

Of course, if you comment out the line that says `include(../build/libusb.pri)` then it won't pick up that package via pkg-config. Could it be that you initially tried to build it without `libusb1-devel` installed, then commented out the `libusb.pri` line, then installed `libusb1-devel`, then found that it didn't find the headers and then fixed up the `#include` lines?

Try a fresh checkout now that you have `libusb1-devel` installed, and it should find all dependencies properly.

BTarn:
Thank you, I tried as you suggested and it's all ok now.
Sorry for taking your time on my silly issue.

thp:
No worries :)

